PR #5, opened 2018-07-02
add --eval option

### LINKED ISSUE #4: Feature request: option for not using ofborg cache
I want to review NixOS/Nixpkgs#42854, but it fails because one of the packages that need to be rebuilt after that PR has since been dropped from Nixpkgs. It would be nice to have the option of not using ofborg cache in such cases.

dotlambda:
Allows to not use ofborg's evaluation result.
Fixes #4.


dotlambda:
Btw, I think nix-review should automatically fallback to local eval if ofborg is currently reevaluating.

Mic92:
Depends, sometimes it can be also annoying if local evaluation has to be performed, while a non-attribute related commit was done.

PR #11, opened 2018-08-30
suspend sigint during cleanup

Mic92:


nix_review/app.py
```diff
@@ -145,25 +146,46 @@ def find_nixpkgs_root() -> Optional[str]:
...
         return self.worktree_dir
 
     def __exit__(self, exc_type: Any, exc_value: Any, traceback: Any) -> None:
-        shutil.rmtree(self.worktree_dir)
-        sh(["git", "worktree", "prune"])
-        os.environ.clear()
-        os.environ.update(self.environ)
+        if self.worktree_dir is None:
+            return
+        with DisableKeyboardInterrupt():
```
Mic92: This can still leave a small window to send a second sigint, but humans are not usually that fast.

PR #49, opened 2019-03-19
Implement support for `nix-review wip` to review uncommited changes

Ma27:
This is the only feature that `nox-review` supports and I'm actually
missing here.

Unstaged (or staged changes with `-s`) will be applied onto the git
workspace in `~/.cache/nix-review/rev-{base hash}-dirty` using `git
apply` and tested after that.

### DIFF
--- a/nix_review/tests/test_wip.py
+++ b/nix_review/tests/test_wip.py
@@ -1,0 +1,71 @@
+import unittest
+
+from typing import Any, List, Tuple
+from unittest.mock import MagicMock, patch
+
+from io import StringIO
+
+from nix_review.cli import main
+
+from .cli_mocks import (
+    CliTestCase,
+    build_cmds,
+    Mock,
+    IgnoreArgument,
+    MockCompletedProcess,
+    read_asset,
+)
+
+
+def wip_command_cmds() -> List[Tuple[Any, Any]]:
+    return [
+        (
+            ["git", "rev-parse", "--verify", "HEAD"],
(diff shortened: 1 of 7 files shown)


Mic92:
Cool. I always wanted to see this feature implemented so I can propose the deprecation of `nox-review`.

nix_review/tests/test_wip.py
```diff
@@ -0,0 +1,71 @@
...
+            MockCompletedProcess(stdout=b"hash1\n"),
+        ),
+        (["git", "worktree", "add", IgnoreArgument, "hash1"], MockCompletedProcess()),
+        (IgnoreArgument, MockCompletedProcess(stdout=StringIO("<items></items>"))),
+        (["git", "apply"], MockCompletedProcess()),
+        (
+            IgnoreArgument,
+            MockCompletedProcess(stdout=StringIO(read_asset("package_list_after.txt"))),
+        ),
+    ]
```
Mic92: You even worked through the nasty-mocked test suite.

Mic92:
Do you know if it possible to derive a new unique hash for every diff? Otherwise
it can easily happen that someone applies a new change, while having the old nix-review still open, which leads to a directory collision.

Mic92:
A different solution would be to append increasing numbers in case of a collision.
This would also solve https://github.com/Mic92/nix-review/issues/47

Mic92:
fixed in https://github.com/Mic92/nix-review/pull/50

Mic92:
Just released 2.0.

PR #57, opened 2019-05-26
Escape build.nix attributes

marsam:
Fixes https://github.com/Mic92/nix-review/issues/56

### DIFF
--- a/nix_review/utils.py
+++ b/nix_review/utils.py
@@ -37,3 +37,10 @@ def verify_commit_hash(commit: str) -> str:
         .stdout.decode("utf-8")
         .strip()
     )
+
+
+def escape_attr(attr: str) -> str:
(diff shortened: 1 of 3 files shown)


nix_review/utils.py
```diff
@@ -37,3 +37,10 @@ def verify_commit_hash(commit: str) -> str:
         .stdout.decode("utf-8")
         .strip()
     )
+
+
+def escape_attr(attr: str) -> str:
+    index = attr.rfind(".")
+    if index == -1:
+        return attr
+    return f'{attr[:index]}."{attr[index+1:]}"'
```
Mic92: In theory probably we should probably split on "." and escape every package attribute, but there is probably no example
in nixpkgs that would require that.

Issue #44, opened 2019-01-25
Can this be used on forks of nixpkgs?


CMCDragonkai:
Or other kinds of nixpkgs repositories like private nixpkgs repos.

Mic92:
In theory yes. In practice nixpkgs is hard-coded at the moment:
https://github.com/Mic92/nix-review/blob/master/nix_review/review.py#L209

Mic92:
This was implemented in https://github.com/Mic92/nixpkgs-review/pull/69

PR #69, opened 2019-10-12
Add `--remote` option to `nix-review rev` for personal forks

Ma27:
Partially addresses #44 (but doesn't fully support private
repositories).

### DIFF
--- a/nix_review/cli/__init__.py
+++ b/nix_review/cli/__init__.py
@@ -62,6 +62,12 @@ def rev_flags(subparsers: argparse._SubParsersAction) -> argparse.ArgumentParser
     rev_parser.add_argument(
         "commit", help="commit/tag/ref/branch in your local git repository"
     )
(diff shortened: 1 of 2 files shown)


nix_review/cli/__init__.py
```diff
@@ -62,6 +62,12 @@ def rev_flags(subparsers: argparse._SubParsersAction) -> argparse.ArgumentParser
     rev_parser.add_argument(
         "commit", help="commit/tag/ref/branch in your local git repository"
     )
+    rev_parser.add_argument(
+        "-r",
+        "--remote",
+        default="https://github.com/NixOS/nixpkgs",
+        help="Name of the nixpkgs repo to review",
+    )
```
Mic92: I think it would also make sense to add this to `wip` as well.

Ma27: You're absolutely right. Fixed.

PR #73, opened 2019-11-04
set nixpkgs config as an attrset instead of a function

veprbl:
The nixpkgs manual seems to prefer attrset and we are not using the argument anyway.

I noticed this due to a recent breakage: https://github.com/NixOS/nixpkgs/pull/72376#discussion_r341883853


Mic92:
fixed in: https://github.com/Mic92/nix-review/pull/74

PR #74, opened 2019-11-04
buildenv: simplify the config

zimbatm:
It's not necessary to use a function since we don't use the pkgs
attribute. The config can just be an attribute set.

Related to https://github.com/NixOS/nixpkgs/pull/72376#discussion_r341883853


Mic92:
Does this also break on current master?

zimbatm:
It was broken for a bit but then my change got reverted (rightfully so).

PR #83, opened 2020-02-16
Add note about -I argument for nix-build

bhipple:
It took me a minute to figure out why the build failed; this may or may not be
worth including in the readme, but it could help someone.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 
 [![Build Status](https://travis-ci.org/Mic92/nixpkgs-review.svg?branch=master)](https://travis-ci.org/Mic92/nixpkgs-review)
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -45,6 +45,15 @@ $ nix-build
 $ ./result/bin/nixpkgs-review
 ```
 
+Note that this asserts formatting with the latest version of
+[black](https://github.com/psf/black), so you may need to specify a more up to
+date version of NixPkgs:
+
+```console
+$ nix-build -I nixpkgs=$HOME/path/to/nixpkgs
```
Mic92: This requires less adaption from the user side:

```suggestion
$ nix-build -I nixpkgs=https://github.com/NixOS/nixpkgs-channels/archive/nixpkgs-unstable.tar.gz
```

PR #84, opened 2020-02-16
Add --no-shell flag for non-interactive nixpkgs-review

bhipple:
Add --no-shell arg for non-interactive nixpkgs-review

Usually we want to interactively review, but in some workflows we may want to
use `nixpkgs-review` to batch process a number of PR reviews.

This commit adds a `--no-shell` flag which, when passed, will do everything
except drop into the interactive `nix-shell`, and instead just finish with an
exit code signaling success or failure.

CC @jonringer @mic92

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,7 +12,7 @@ NOTE: this project used to be called `nix-review`
 - [ofborg](https://github.com/NixOS/ofborg) support: reuses evaluation output of CI to skip local evaluation, but
   also fallbacks if ofborg is not finished
 - automatically detects target branch of pull request
-- provides a `nix-shell` with all packages, that did not fail to build
+- provides a `nix-shell` with all packages that did not fail to build
 - remote builder support
 - allows to build a subset of packages (great for mass-rebuilds)
 - allow to build nixos tests
@@ -127,6 +127,39 @@ Staged changes can be reviewed like this:
 $ nixpkgs-review wip --staged
 ```
 
+## Using nix-review in scripts
+
+After building, `nixpkgs-review` will normally start a `nix-shell` with the
+packages built, to allow for interactive testing. To use `nixpkgs-review`
+non-interactively in scripts, use the `--no-shell` command, which can allow for
+batch processing of multiple reviews or use in scripts/bots.
+
+Example testing multiple unrelated PRs:
+
+```bash
+for pr in 807{60..70}; do
+    nixpkgs-review pr --no-shell $pr && echo "PR $pr succeeded" || echo "PR $pr failed"
+done
+```
+
+Example composing `nixpkgs-review` inside a script and using a token from
+[hub](https://github.com/github/hub) to post a comment on the PR with results:
+
+```
+review_pr_with_comment() {
+    nixpkgs-review pr --no-shell $1
+    escaped_msg=$(sed -e 's/"/\\"/g' ${XDG_CACHE_HOME:-${HOME}/.cache}/nixpkgs-review/pr-$1/report.md)
+    TOKEN=$(awk '/oauth_token/ {print $NF}' ~/.config/hub)
+    curl -sH "Authorization: token $TOKEN" -XPOST \
+        -d "{\"body\": \"$msg\"}" https://api.github.com/repos/NixOS/nixpkgs/issues/$1/comments
+}
+
+review_pr_with_comment 80767
+```
+
+The above is just a basic example for ideas.
+
+
 ## Remote builder:
 
 Nix-review will pass all arguments given in `--build-arg` to `nix-build`:
@@ -214,7 +247,6 @@ Note that this has been not yet implemented for pull requests i.e. `pr` subcomma
 
 ## Roadmap
 
-- [ ] trigger ofBorg builds (write @GrahamcOfBorg build foo into pull request discussion)
 - [ ] build on multiple platforms
 - [ ] test backports
 - [ ] show pull request description + diff during review
--- a/README.md
+++ b/README.md
@@ -151,7 +151,7 @@ review_pr_with_comment() {
     escaped_msg=$(sed -e 's/"/\\"/g' ${XDG_CACHE_HOME:-${HOME}/.cache}/nixpkgs-review/pr-$1/report.md)
     TOKEN=$(awk '/oauth_token/ {print $NF}' ~/.config/hub)
     curl -sH "Authorization: token $TOKEN" -XPOST \
-        -d "{\"body\": \"$msg\"}" https://api.github.com/repos/NixOS/nixpkgs/issues/$1/comments
+        -d "{\"body\": \"$escaped_msg\"}" https://api.github.com/repos/NixOS/nixpkgs/issues/$1/comments
 }
 
 review_pr_with_comment 80767
--- a/nixpkgs_review/cli/__init__.py
+++ b/nixpkgs_review/cli/__init__.py
@@ -138,6 +138,11 @@ def parse_args(command: str, args: List[str]) -> argparse.Namespace:
             type=regex_type,
             help="Regular expression that package attributes have to match (can be passed multiple times)",
         ),
+        CommonFlag(
+            "--no-shell",
+            action="store_true",
+            help="Only evaluate and build without executing nix-shell",
+        ),
     ]
 
     for flag in common_flags:
--- a/nixpkgs_review/cli/pr.py
+++ b/nixpkgs_review/cli/pr.py
@@ -45,6 +45,7 @@ def pr_command(args: argparse.Namespace) -> None:
                 review = Review(
                     builddir=builddir,
                     build_args=args.build_args,
+                    no_shell=args.no_shell,
                     api_token=args.token,
                     use_ofborg_eval=use_ofborg_eval,
                     only_packages=set(args.package),
--- a/nixpkgs_review/review.py
+++ b/nixpkgs_review/review.py
@@ -79,6 +79,7 @@ class Review:
         self,
         builddir: Builddir,
         build_args: str,
+        no_shell: bool,
         api_token: Optional[str] = None,
         use_ofborg_eval: Optional[bool] = True,
         only_packages: Set[str] = set(),
@@ -87,6 +88,7 @@ class Review:
     ) -> None:
         self.builddir = builddir
         self.build_args = build_args
+        self.no_shell = no_shell
         self.github_client = GithubClient(api_token)
         self.use_ofborg_eval = use_ofborg_eval
         self.checkout = checkout
@@ -186,7 +188,10 @@ class Review:
         report = Report(attr)
         report.print_console(pr)
         report.write(self.builddir.path, pr)
-        nix_shell(report.built_packages(), self.builddir.path)
+        if self.no_shell:
+            sys.exit(0 if report.succeeded() else 1)
+        else:
+            nix_shell(report.built_packages(), self.builddir.path)
 
     def review_commit(
         self,
@@ -357,6 +362,7 @@ def review_local_revision(
         review = Review(
             builddir=builddir,
             build_args=args.build_args,
+            no_shell=args.no_shell,
             only_packages=set(args.package),
             package_regexes=args.package_regex,
         )
--- a/nixpkgs_review/report.py
+++ b/nixpkgs_review/report.py
@@ -100,6 +100,10 @@ class Report:
         self.write_markdown(directory, pr)
         write_error_logs(self.attrs, directory)
 
+    def succeeded(self) -> bool:
+        """Whether the report is considered a success or a failure"""
+        return len(self.failed) == 0
+
     def write_markdown(self, directory: Path, pr: Optional[int]) -> None:
         with open(directory.joinpath("report.md"), "w+") as f:
             cmd = "nixpkgs-review"


bhipple:
Example of a PR that I'm using this for:
https://github.com/NixOS/nixpkgs/pull/80301

I was just running `nix-build -A <attr>`, but `nixpkgs-review wip` will catch a lot more.  However, since I'm scripting it across ~200 packages I really just want to see if the pkg builds/tests cleanly along with its dependencies.  Using this feature I can do that :)

jonringer:
another option would be `--no-shell` to avoid dropping into a shell. Just exit with a non-zero return if something failed.

bhipple:
I started out with `--non-interactive`, but then when looking at logic like `if not non_interactive then nix-shell` thought maybe avoiding the double negative would be preferable.  I do like how `--no-shell` is very explicit about exactly what it's doing, though.

jonringer:
I would see what @Mic92 thinks

bhipple:
First implementation didn't actually propagate the exit code, slightly defeating the purpose :sweat_smile: Now it will actually exit non-zero when there are > 0 failures.

```
λ brh nixpkgs →  ~/src/nixpkgs-review/result/bin/nixpkgs-review wip --batch
$ git -c fetch.prune=false fetch --force https://github.com/NixOS/nixpkgs master:refs/nixpkgs-review/0
$ git worktree add /home/bhipple/.cache/nixpkgs-review/rev-72155667aeb038b54b07304afd9030ab2e2f22e5-dirty-1/nixpkgs 72155667aeb038b54b07304afd9030ab2e2f22e5
Preparing /home/bhipple/.cache/nixpkgs-review/rev-72155667aeb038b54b07304afd9030ab2e2f22e5-dirty-1/nixpkgs (identifier nixpkgs1)
HEAD is now at 72155667aeb Merge pull request #79908 from bkchr/bkchr-fix-kde-pim-runtime
$ nix-env -f /home/bhipple/.cache/nixpkgs-review/rev-72155667aeb038b54b07304afd9030ab2e2f22e5-dirty-1/nixpkgs -qaP --xml --out-path --show-trace
Applying `nixpkgs` diff...
$ nix-env -f /home/bhipple/.cache/nixpkgs-review/rev-72155667aeb038b54b07304afd9030ab2e2f22e5-dirty-1/nixpkgs -qaP --xml --out-path --show-trace --meta
2 package updated:
railcar rainicorn

$ nix build --no-link --keep-going --option build-use-sandbox relaxed -f /home/bhipple/.cache/nixpkgs-review/rev-72155667aeb038b54b07304afd9030ab2e2f22e5-dirty-1/build.nix
builder for '/nix/store/vmkxx7bjgr7cdjhvymxhjqxhxbcf77ii-railcar-1.0.4.drv' failed with exit code 1; last 10 log lines:

  ERROR: cargoSha256 is out of date

  Cargo.lock is not the same in railcar-1.0.4-vendor.tar.gz

  To fix the issue:
  1. Use "1111111111111111111111111111111111111111111111111111" as the cargoSha256 value
  2. Build the derivation and wait it to fail with a hash mismatch
  3. Copy the 'got: sha256:' value back into the cargoSha256 field
cannot build derivation '/nix/store/x32rbv8smmdd73w1igl43vwwrx834hlq-env.drv': 1 dependencies couldn't be built
[0 built (1 failed)]
error: build of '/nix/store/x32rbv8smmdd73w1igl43vwwrx834hlq-env.drv' failed
1 package marked as broken and skipped:
rainicorn

1 package failed to build:
railcar

$ git worktree prune

λ brh nixpkgs -1-  →  echo $?
1
```

jonringer:
you can also do:
```
 ~/src/nixpkgs-review/result/bin/nixpkgs-review wip --batch && echo succeeded  || echo failed
```

nixpkgs_review/review.py
```diff
@@ -186,7 +188,10 @@ def start_review(self, attr: List[Attr], pr: Optional[int] = None) -> None:
         report = Report(attr)
         report.print_console(pr)
         report.write(self.builddir.path, pr)
-        nix_shell(report.built_packages(), self.builddir.path)
+        if self.batch:
+            sys.exit(0 if report.succeeded() else 1)
+        else:
+            nix_shell(report.built_packages(), self.builddir.path)
```
Mic92: To make this more general, I would instead propose to expose the `--run` argument of
`nix-shell`:

```
nix_shell(report.built_packages(), self.builddir.path, run=self.run)
```

This would also allow to interact with build artifacts more easily (they are stored in ./results.) before they are removed by nixpkgs-review again.
The exit status of `nix-shell` is also exposed to the user (when the command fail, python will also fail).

Mic92: I have not thought about on how to expose the build status to the command.
An additional environment variable within the nix-shell expression might help?

jonringer: I think that deviates slightly with the intention of the original PR. @bhipple just wants a command to tell him "Yes, this doesn't cause any regressions" or "No, theses now fail". 

I guess `--batch`,`--run args` could be a mutually exclusive switch. But --run still sounds like a different feature altogether.

I still think there's value in being able to have nix-review give you a "quick" yes/no. Also would be useful in CI applications

Mic92: Well in that case you would just do `--run true`.

jonringer: I would still say that it's less intuitive, similar to how you used to have to do `buildPhase = ":";` instead of `doBuild = false;` in nixpkgs

Mic92: Ok. Let's rename it to no-shell instead.

README.md
```diff
@@ -12,7 +12,8 @@ NOTE: this project used to be called `nix-review`
 - [ofborg](https://github.com/NixOS/ofborg) support: reuses evaluation output of CI to skip local evaluation, but
   also fallbacks if ofborg is not finished
 - automatically detects target branch of pull request
-- provides a `nix-shell` with all packages, that did not fail to build
+- provides a `nix-shell` with all packages that did not fail to build
+- provides a `--batch` mode to run `nixpkgs-review` non-interactively for scripted verification
```
Mic92: Please move this to its own section called `Non-interactive execution` with an example.

nixpkgs_review/cli/__init__.py
```diff
@@ -138,6 +138,11 @@ def parse_args(command: str, args: List[str]) -> argparse.Namespace:
             type=regex_type,
             help="Regular expression that package attributes have to match (can be passed multiple times)",
         ),
+        CommonFlag(
+            "--batch",
```
Mic92: Please move this to common_flags and also make it work for the `rev` command.
Also rename it to `--no-shell`.

nixpkgs_review/cli/__init__.py
```diff
@@ -138,6 +138,11 @@ def parse_args(command: str, args: List[str]) -> argparse.Namespace:
...
+            action="store_true",
+            help="Batch process without stopping to execute interactive commands like nix-shell (for scripted builders)",
```
Mic92:
```suggestion
            help="Only evaluate and build without executing nix-shell",
```

nixpkgs_review/cli/pr.py
```diff
@@ -45,6 +45,7 @@ def pr_command(args: argparse.Namespace) -> None:
                 review = Review(
                     builddir=builddir,
                     build_args=args.build_args,
+                    batch=args.batch,
```
Mic92:
```suggestion
                    no_shell=args.no_shell,
```

nixpkgs_review/review.py
```diff
@@ -79,6 +79,7 @@ def __init__(
         self,
         builddir: Builddir,
         build_args: str,
+        batch: bool,
```
Mic92:
```suggestion
        no_shell: bool,
```

nixpkgs_review/review.py
```diff
@@ -87,6 +88,7 @@ def __init__(
     ) -> None:
         self.builddir = builddir
         self.build_args = build_args
+        self.batch = batch
```
Mic92:
```suggestion
        self.no_shell = no_shell
```

nixpkgs_review/review.py
```diff
@@ -186,7 +188,10 @@ def start_review(self, attr: List[Attr], pr: Optional[int] = None) -> None:
         report = Report(attr)
         report.print_console(pr)
         report.write(self.builddir.path, pr)
-        nix_shell(report.built_packages(), self.builddir.path)
+        if self.batch:
```
Mic92:
```suggestion
        if self.no_shell:
```

nixpkgs_review/review.py
```diff
@@ -357,6 +362,7 @@ def review_local_revision(
         review = Review(
             builddir=builddir,
             build_args=args.build_args,
+            batch=args.batch,
```
Mic92:
```suggestion
            no_shell=args.no_shell,
```

nixpkgs_review/cli/__init__.py
bhipple: Tested working for all 3 commands; it's also already in common_flags.

README.md
bhipple: Good idea; I wrote up some basic examples in the documentation for inspiration.

README.md
```diff
@@ -214,7 +251,6 @@ Note that this has been not yet implemented for pull requests i.e. `pr` subcomma
 
 ## Roadmap
 
-- [ ] trigger ofBorg builds (write @GrahamcOfBorg build foo into pull request discussion)
```
bhipple: Per comment from @grahamc, commenting explicitly for the OfBorg bot is now unnecessary, so I've removed this item from the roadmap.

README.md
```diff
@@ -127,6 +127,43 @@ Staged changes can be reviewed like this:
...
+After building, `nixpkgs-review` will normally start a `nix-shell` with the
+packages built, to allow for interactive testing. To use `nixpkgs-review`
+non-interactively in scripts, use the `--no-shell` command, which can allow for
+batch processing of multiple reviews or use in scripts/bots.
+
+Example testing multiple unrelated PRs:
+
+```bash
+for pr in 807{60..70}; do
+    nixpkgs-review pr --no-shell $pr && echo "PR $pr succeeded" || echo "PR $pr failed"
```
Mic92: Actually you can even do `nixpkgs-review pr 807{60..70}`

README.md
```diff
@@ -127,6 +127,43 @@ Staged changes can be reviewed like this:
...
+
+Example composing `nixpkgs-review` inside a script and using a token from
+[hub](https://github.com/github/hub) to post a comment on the PR with results:
+
+```
+review_pr_with_comment() {
+    if ~/src/nixpkgs-review/result/bin/nixpkgs-review pr --no-shell $1; then
+        msg="nixpkgs-review succeeded"
+    else
+        msg="nixpkgs-review failed"
```
Mic92: What do you think about using?

msg=$(cat ${XDG_CACHE_HOME:-HOME}/.cache/nix-review/pr-$1/report.md)

README.md
```diff
@@ -127,6 +127,43 @@ Staged changes can be reviewed like this:
...
+    fi
+
+    TOKEN=$(grep oauth_token ~/.config/hub | awk '{print $NF}')
```
Mic92:
```suggestion
    TOKEN=$(awk '/oauth_token/ {print $NF}' ~/.config/hub)
```

README.md
bhipple: I think that's a great idea; that looks a lot more full featured than the bare placeholder example comment I was putting here.

That said, I can't quite get the quoting in the `$msg` to work ... do you know how to sanitize it before giving to curl?

README.md
```diff
@@ -147,12 +147,8 @@ Example composing `nixpkgs-review` inside a script and using a token from
...
 ```
 review_pr_with_comment() {
-    if ~/src/nixpkgs-review/result/bin/nixpkgs-review pr --no-shell $1; then
-        msg="nixpkgs-review succeeded"
-    else
-        msg="nixpkgs-review failed"
-    fi
-
+    ~/src/nixpkgs-review/result/bin/nixpkgs-review pr --no-shell $1
+    msg=$(cat ${XDG_CACHE_HOME:-${HOME}/.cache}/nixpkgs-review/pr-$1/report.md)
```
Mic92: This should be sufficient for escaping inside json:

```suggestion
    escaped_msg=$(sed -e 's/"/\\"/g' ${XDG_CACHE_HOME:-${HOME}/.cache}/nixpkgs-review/pr-$1/report.md)
```

README.md
```diff
@@ -127,6 +127,39 @@ Staged changes can be reviewed like this:
...
+    nixpkgs-review pr --no-shell $pr && echo "PR $pr succeeded" || echo "PR $pr failed"
+done
+```
+
+Example composing `nixpkgs-review` inside a script and using a token from
+[hub](https://github.com/github/hub) to post a comment on the PR with results:
+
+```
+review_pr_with_comment() {
+    ~/src/nixpkgs-review/result/bin/nixpkgs-review pr --no-shell $1
```
Mic92:
```suggestion
    nixpkgs-review pr --no-shell $1
```

bhipple: Oops, thanks; I was obviously testing with my local build :)

README.md
bhipple: That wouldn't get the individual follow-up command on each PR, though, would it?  The minimalist example is just to show here that you can "run a nixpkgs-review and then some cmd, based on whether it succeeded or failed for that particular review".  The `echo` is a bit of a placeholder here.

README.md
```diff
@@ -127,6 +127,39 @@ Staged changes can be reviewed like this:
...
+Example composing `nixpkgs-review` inside a script and using a token from
+[hub](https://github.com/github/hub) to post a comment on the PR with results:
+
+```
+review_pr_with_comment() {
+    nixpkgs-review pr --no-shell $1
+    escaped_msg=$(sed -e 's/"/\\"/g' ${XDG_CACHE_HOME:-${HOME}/.cache}/nixpkgs-review/pr-$1/report.md)
+    TOKEN=$(awk '/oauth_token/ {print $NF}' ~/.config/hub)
+    curl -sH "Authorization: token $TOKEN" -XPOST \
+        -d "{\"body\": \"$msg\"}" https://api.github.com/repos/NixOS/nixpkgs/issues/$1/comments
```
Mic92:
```suggestion
        -d "{\"body\": \"$escaped_msg\"}" https://api.github.com/repos/NixOS/nixpkgs/issues/$1/comments
```

bhipple:
Your changes LGTM, thanks @Mic92; any further sugestions or are we good to go here?

Issue #89, opened 2020-03-29
nixpkgs-review pr <num> --post-result --approve --merge flags


bhipple:
Often, code reviewers will run `nixpkgs-review pr <num>` then manually paste the
results into the code review, like this:

https://github.com/NixOS/nixpkgs/pull/83695

I would like to add the ability to provide a `nixpkgs-review pr 1234 --comment` flag which will do this automatically, using the markdown report that `nixpkgs-review` is already producing.

Mic92:
Sounds good to me.

FRidh:
Maybe call it `--post-result` instead of `--comment`. When I saw `--comment` I had at least the impression I could add additional text. Note that I don't really mind.

bhipple:
Having a `--comment` that lets you add additional text seems like a nice idea, actually.  Something like this is where I'd like to get eventually:

```bash
nixpkgs-review pr 1234 --no-shell --post-result --approve --merge --comment "LGTM, thanks for the contribution!"
```

[I basically have this already](https://github.com/bhipple/dotfiles/blob/master/bin/borg), and @jonringer and I have been using it on PRs for the past week or so and love it, so I'm looking to upstream the features into `nixpkgs-review` for wider consumption.

jonringer:
I agree with @FRidh , I thought it would take another argument as well, allowing for something like `--comment "LGTM\nhas tests :+1:". --post-result is more explicit.

Mic92:
This is now available in NUR as a preview: `nur.repos.mic92.nixpkgs-review-unstable`

Mic92:
This is now released: https://github.com/Mic92/nixpkgs-review/tree/2.3.0

NickHu:
Sorry for flogging a dead horse, but is there a way to do attach a comment to `nixpkgs-review post-result` as @jonringer suggested in his last comment? I saw something like that in one of the PRs he merged, but I wasn't able to figure out how to do it from the `--help` pages.

Mic92:
No. Not at the moment, but it is not hard to add it.

PR #90, opened 2020-03-29
Refactor the report.markdown() function

bhipple:
This commit has no functional changes; it merely refactors the
`Report.markdown()` function to be a pure function returning a `str`, so
that we can thread it through to a `request` post in a future update as
part of #89.

Along the way, this adds test coverage to the function to help catch
refactoring bugs, issues like #88, and make the final report output format
easier to view (e.g., can just open the test file to see what it looks like
when reviewing changes to the format).

### DIFF
--- a/nixpkgs_review/report.py
+++ b/nixpkgs_review/report.py
@@ -1,7 +1,7 @@
 import os
 import subprocess
 from pathlib import Path
-from typing import IO, Callable, List, Optional
+from typing import Callable, List, Optional
 
 from .nix import Attr
 from .utils import info, warn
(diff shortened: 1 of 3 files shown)


bhipple:
The first commit in this PR is https://github.com/Mic92/nixpkgs-review/pull/88 and will drop off when that one is merged, but we need both commits to pass the test suite now :) 

CC @Mic92 @jonringer @ryantm

Mic92:
It is interesting though that the github action for testing is no longer triggered.

Mic92:
The problem is that `Report` has side-effects:

```
======================================================================
ERROR: test_markdown_report (nixpkgs_review.tests.test_report.ReportTestcase)
Test that the markdown report format is as expected
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/build/nixpkgs-review/nixpkgs_review/tests/test_report.py", line 31, in test_markdown_report
    report = Report([foo, bar, baz])
  File "/build/nixpkgs-review/nixpkgs_review/report.py", line 90, in __init__
    elif not a.was_build():
  File "/build/nixpkgs-review/nixpkgs_review/nix.py", line 32, in was_build
    ["nix-store", "--verify-path", self.path], stderr=subprocess.DEVNULL
  File "/nix/store/xb1mp0f16lv23rwhfdc1665v5r20jfmm-python3-3.7.6/lib/python3.7/subprocess.py", line 488, in run
    with Popen(*popenargs, **kwargs) as process:
  File "/nix/store/xb1mp0f16lv23rwhfdc1665v5r20jfmm-python3-3.7.6/lib/python3.7/subprocess.py", line 800, in __init__
    restore_signals, start_new_session)
  File "/nix/store/xb1mp0f16lv23rwhfdc1665v5r20jfmm-python3-3.7.6/lib/python3.7/subprocess.py", line 1551, in _execute_child
    raise child_exception_type(errno_num, err_msg, err_filename)
FileNotFoundError: [Errno 2] No such file or directory: 'nix-store': 'nix-store'
```

bhipple:
Was a simple fix; the mock `Attr` needs to always have the "was built" boolean set, otherwise it tries to build it.  Now passes `nix-build` locally for me.

PR #93, opened 2020-04-04
Implement --post-result function for PR review comments

### LINKED ISSUE #89: nixpkgs-review pr <num> --post-result --approve --merge flags

bhipple:
This extends `nixpkgs-review` with a `--comment` flag, active during the `pr`
subcommand, that will post a comment on the PR review with the results.

Resolves #89

CC @mic92 @jonringer @ryantm @fridh

### DIFF
--- a/README.md
+++ b/README.md
@@ -109,10 +109,10 @@ redis-cli 4.0.8
 ```
 
 If you'd like to post the `nixpkgs-review` results as a formatted PR comment,
-pass the `--comment` flag:
+pass the `--post-result` flag:
 
 ```console
-$ nixpkgs-review pr --comment 37242
+$ nixpkgs-review pr --post-result 37242
 ```
 
 To review a local commit without pull request, use the following command:
@@ -147,7 +147,7 @@ comments for later review:
 
 ```bash
 for pr in 807{60..70}; do
-    nixpkgs-review pr --no-shell --comment $pr && echo "PR $pr succeeded" || echo "PR $pr failed"
+    nixpkgs-review pr --no-shell --post-result $pr && echo "PR $pr succeeded" || echo "PR $pr failed"
 done
 ```
 
@@ -165,7 +165,7 @@ This allows to parallelize builds across multiple machines.
 
 ## Github api token
 
-The `pr --comment` option requires a Github API token, and even for read-only
+The `pr --post-result` option requires a Github API token, and even for read-only
 calls github returns 403 error messages if your IP hits the rate limit for
 unauthenticated calls.
 
--- a/README.md
+++ b/README.md
@@ -165,16 +165,17 @@ This allows to parallelize builds across multiple machines.
 
 ## Github api token
 
-In case your IP exceeds the rate limit, github will return an 403 error message.
-To increase your limit first create a [personal access token](https://help.github.com/articles/creating-a-personal-access-token-for-the-command-line/).
-Then use either the `--token` parameter of the `pr` subcommand or
-set the `GITHUB_OAUTH_TOKEN` environment variable.
+The `pr --comment` option requires a Github API token, and even for read-only
+calls github returns 403 error messages if your IP hits the rate limit for
+unauthenticated calls.
+
+To use a token, first create a [personal access token](https://help.github.com/articles/creating-a-personal-access-token-for-the-command-line/).
+Then use either the `--token` parameter of the `pr` subcommand or set the
+`GITHUB_OAUTH_TOKEN` environment variable.
 
 ```console
 $ nixpkgs-review pr --token "5ae04810f1e9f17c3297ee4c9e25f3ac1f437c26" 37244
 ```
-
-This is required for the `pr --comment` option.
 
 ## Checkout strategy (recommend for r-ryantm + cachix)
 
--- a/README.md
+++ b/README.md
@@ -18,6 +18,7 @@ NOTE: this project used to be called `nix-review`
 - allow to build nixos tests
 - colorful output
 - markdown reports
+- GitHub integration to post PR comments with results
 - logs per built or failed package
 - symlinks build packages to result directory for inspection
 
@@ -107,6 +108,13 @@ $ nix-shell -p redis
 redis-cli 4.0.8
 ```
 
+If you'd like to post the `nixpkgs-review` results as a formatted PR comment,
+pass the `--comment` flag:
+
+```console
+$ nixpkgs-review pr --comment 37242
+```
+
 To review a local commit without pull request, use the following command:
 
 ```console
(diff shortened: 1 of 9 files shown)


bhipple:
Note that even though I'm bumping PyGithub to 1.47, that's an unrelated change; this works just fine with older versions.  I was bumping it because I saw newer versions added additional `mypy` types, and was hoping it'd resolve the `pyi` file issue stated above, but it didn't.

jonringer:
don't you have to generate a github token?

If so, then you should also include the steps to do so

nixpkgs_review/github.py
```diff
@@ -16,6 +24,11 @@ def get(self, path: str) -> Any:
             req.add_header("Authorization", f"token {self.api_token}")
         return json.loads(urllib.request.urlopen(req).read())
 
+    def pr_comment(self, pr: int, msg: str) -> None:
+        "Post a comment on a PR with nixpkgs-review report"
+        gh_pr = self.nixpkgs.get_pull(pr)
+        gh_pr.create_issue_comment(msg)
```
Mic92, line 30: If you just create a simple comment, I would rather prefer to create a HTTP post request using the API directly: https://developer.github.com/v3/issues/comments/

Than we don't need to add external library.

nixpkgs_review/github.py
```diff
@@ -1,13 +1,21 @@
...
-from collections import defaultdict
+import urllib.parse
 import urllib.request
+
+from collections import defaultdict
 from typing import Any, DefaultDict, Dict, Optional, Set
 
+# The PyGithub module does have stub files, but they're not installed via its
+# setup.py, so mypy complains about missing types.
+from github import Github  # type: ignore
```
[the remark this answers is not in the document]
Mic92: see https://github.com/NixOS/nixpkgs/pull/82453 regarding mypy in NixOS.

nixpkgs_review/github.py
Mic92: One can use https://docs.python.org/3/library/http.client.html to make POST requests to github.

nixpkgs_review/github.py
```diff
@@ -16,6 +24,11 @@ def get(self, path: str) -> Any:
             req.add_header("Authorization", f"token {self.api_token}")
         return json.loads(urllib.request.urlopen(req).read())
```
Mic92, line 25: I would suggest to move the code above to `_request(..., method)` and add `Request(url, method=method)`. Then you can introduce a function called `get()` that calls `_request(..., method="GET")` and one for `post()` that calls `_request(..., method="POST")`

Mic92:
> don't you have to generate a github token?
> 
> If so, then you should also include the steps to do so

We already do as part of the readme.

README.md
```diff
@@ -174,10 +165,13 @@ This allows to parallelize builds across multiple machines.
...
-In case your IP exceeds the rate limit, github will return an 403 error message.
-To increase your limit first create a [personal access token](https://help.github.com/articles/creating-a-personal-access-token-for-the-command-line/).
-Then use either the `--token` parameter of the `pr` subcommand or
-set the `GITHUB_OAUTH_TOKEN` environment variable.
+The `pr --comment` option requires a Github API token, and even for read-only
+calls github returns 403 error messages if your IP hits the rate limit for
+unauthenticated calls.
+
+To use a token, first create a [personal access token](https://help.github.com/articles/creating-a-personal-access-token-for-the-command-line/).
+Then use either the `--token` parameter of the `pr` subcommand or set the
```
Mic92, line 173: Please add a check in `pr_command` to verify if token was set if the comment option was passed.

Mic92: If it is not set the error message should contain a link to this section of the README.

nixpkgs_review/github.py
bhipple: I actually spent many hours wrestling with the JSON encoding prior to going down this route.  In particular, the payload for the message requires packing it into a JSON structure like this:
https://github.com/bhipple/dotfiles/blob/master/bin/borg#L20-L25

Something about packing the string `msg` into `{ "data": msg }` made GitHub really unhappy, but they don't give you any error message; just an error code with a msg saying invalid JSON.  Using the library is nice because it handles all the details of that structure for you.

In a follow-up, I'm also planning on adding `nixpkgs-reivew pr 1234 --comment --approve --merge`, which can approve (anyone) and also merge (if token has maintainer permission), just like the script I linked above.  

Since I'm going to hit 3+ endpoints and the JSON encoding is weird (maybe you know how to do it?), and since the library appears to be well-maintained and even actively supports mypy typing everywhere, I think it's reasonable to take a dependency on it rather than re-implementing the requests ourselves with the low-level `urllib`.  What do you think?

bhipple: With the library, the others look like this:

```python
    def pr_comment(self, pr: int, msg: str) -> None:
        "Post a comment on a PR with nixpkgs-review report"
        gh_pr = self.nixpkgs.get_pull(pr)
        gh_pr.create_issue_comment(msg)

    def pr_approve(self, pr: int) -> None:
        gh_pr = self.nixpkgs.get_pull(pr)
        gh_pr.create_review(event="APPROVE")

    def pr_merge(self, pr: int) -> None:
        gh_pr = self.nixpkgs.get_pull(pr)
        gh_pr.merge()
```
Seems nice and simple/clean.  I just have the comment in this PR since I figured it has more straightforward semantics and would be easier to review/merge/integrate first, but if you want I can update the PR to send all 3 for code review at once.

bhipple: Hmm weird, now I'm getting import errors due to the external lib ... can't win :roll_eyes: 

This was working yesterday, but before I dig in too deeply to fix, do you want to use the lib or rewrite in terms of the lower-level urllib?

Mic92: I can have a look later on. Not having a dependency makes future maintenance easier for me.

nixpkgs_review/github.py
Mic92: Done

nixpkgs_review/github.py
Mic92: Merging and approving should not be flags of the `pr` command in my opinion but rather a subcommand that can be called from within the `nix-shell`.
Otherwise it would mean you would approve/merge a PR regardless of the review done in nixpkgs-review.

i.e.

```
$ nixpkgs-review pr 84428
....
nix-shell> nixpkgs-review merge
```

or

```
$ nixpkgs-review pr 84428
....
nix-shell> nixpkgs-review approve
```

The current pull request number could be passed over environment variables.

PR #117, opened 2020-07-08
Imrpove report package formatting + generic command checker

blaggacao:
Generic command checker implements the idea of: https://github.com/NixOS/nixpkgs/pull/92676/files#diff-a6e8ff6324aa0356e91850bdb64f28b6R54


nixpkgs_review/tests/test_report.py
```diff
@@ -23,9 +23,9 @@ def mkAttr(name: str, success: bool) -> Attr:
 class ReportTestcase(unittest.TestCase):
     def test_markdown_report(self) -> None:
         "Test that the markdown report format is as expected"
-        foo = mkAttr("foo", True)
-        bar = mkAttr("bar", True)
-        baz = mkAttr("baz", False)
+        foo = mkAttr("foo", True, "bash")
+        bar = mkAttr("bar", True, "sh")
```
blaggacao, line 27: A generally available utility which does _not_ support the `--help` flag? (to test the failing case properly)

nixpkgs_review/report.py
```diff
@@ -39,6 +40,21 @@ def html_pkgs_section(packages: List[Attr], msg: str, what: str = "package") ->
     return res
 
 
+def md_pkgs_generic_test(packages: List[Attr]) -> str:
+    if len(packages) == 0:
+        return ""
+    res = "\n"
+    for pkg in packages:
+        if subprocess.call([pkg.path, "--help"],
```
Mic92, line 48: This sounds like good idea, however:

- It's actually a bit scary to run arbitrary programs. I know this was done by https://github.com/ryantm/nixpkgs-update however this is a dedicated account not having access to the user's home. 
- Some programs might also hang indefinitely...

We could rely on the build sandbox if available to limit the scope but also in this case a timeout is needed (I think nix-build supports that).

blaggacao: - [ ] I just added the timeout
- [ ] Could you guide me on how to ensure this is only done in sandboxes?
- [ ] Please add a commit to this fixing the mocks for the tests to go green, I've tried it for nearly 5 hours with no success. :pray:

Mic92: I think the cleanest way would be to do it as part of the normal build process. During the build we write a `build.nix`. This needs to be extended to have derivation like:

```nix
{ pkgs }:
let
  checkCommands = pkgs.callPackage "/absolute/path/to/nixpkgs_review/nixpkgs_review/nix/checkCommands.nix" {};
in {
  helloCheck = runCommand "check-hello" {} ''
    ${checkCommands}/bin/checkCommands ${hello} $out/report.json
  '';
}
```

`checkCommands` could be a python script that does the subprocess stuff that you are already doing and than write a json file. You can use `nix eval` to get derivation paths out of `build.nix` and than read the json from there. Does this make sense? It is not as straight forward as the current implementation however I don't want to execute arbitrary code on the reviewers private machine.

Mic92: If you write the code I can take care of the tests afterwards.

Mic92: Another thing that is interesting is to catch segfaults. subprocess should have some exception in case signals such as SIGSEV are thrown.

blaggacao:
@Mic92 I din't reach as far as I wished to.

- Manually validated: https://github.com/Mic92/nixpkgs-review/pull/117/files#diff-6aa0e51e7a803c4e75e355d68a1ba117
- Sharp-eye (non-)validated: https://github.com/Mic92/nixpkgs-review/pull/117/files#diff-bc1587e41fc754620495ca73e524214f
- Based on the manually validated json: https://github.com/Mic92/nixpkgs-review/pull/117/files#diff-6bf4dcd62dcb938f8ab41c5a7239e665

```json
[
    {
        "attribute": "cat",
        "result": "succeded"
    },
    {
        "attribute": "bash",
        "result": "timed out"
    },
]
```

nixpkgs_review/report.py
blaggacao: The idea with SIGSEV isn't exposed by `timout`. I didn't want to use python, since it was already sufficently omplex to get my head arround the way this derivation is expected to be written. I figured with bash it would be easier for a start.

Mic92:
I will make nixpkgs-review nix flakes ready over the next time.

Mic92:
With nix flakes you have to use `--expr`:

```console
$ nix eval --json  --impure --expr ./nix/evalAttrs.nix
```

blaggacao:
Can you lift me over this obstacle? I've still very little idea how this is supposed to be working.
I'd be happy to finish this PR off (pre-test)

Mic92:
From your explanation it is not clear to me, what your problem is.

blaggacao:
I couldn't conceive how to populate the `report` struct accessed [here](https://github.com/Mic92/nixpkgs-review/pull/117/files#diff-6bf4dcd62dcb938f8ab41c5a7239e665R47-R56) from the json file. Unfortunately, it's not obvious to me, nor did I achieve to produce the insight I would have needed.

PR #119, opened 2020-07-22
add git to wrapper dependencies

ryantm:


default.nix
```diff
@@ -18,6 +18,7 @@ python3.pkgs.buildPythonApplication rec {
   '';
   makeWrapperArgs = [
     "--prefix PATH" ":" "${nix}/bin"
+    "--prefix PATH" ":" "${git}/bin"
```
Mic92, line 21:
```suggestion
    --prefix PATH : ${stdenv.lib.makeBinPath [ nix git ]}
```

default.nix
```diff
@@ -17,7 +17,7 @@ python3.pkgs.buildPythonApplication rec {
     mypy --strict nixpkgs_review
   '';
   makeWrapperArgs = [
-    "--prefix PATH" ":" "${nix}/bin"
+    --prefix PATH : ${stdenv.lib.makeBinPath [ nix git ]}
```
Mic92, line 20: Sorry. I forget the quotes:

```suggestion
    "--prefix" "PATH" ":" "${stdenv.lib.makeBinPath [ nix git ]}"
```

ryantm:
@Mic92 fixed and squashed.

ryantm:
Oh, maybe I'm not using enough quotes. Why do you use so many?

Mic92:
> Oh, maybe I'm not using enough quotes. Why do you use so many?

I was not sure how it would be passed to make wrapper but apparently not quoted. Otherwise it would just receive one but argument.

PR #142, opened 2020-10-28
Add platform run on to report

### LINKED ISSUE #139: Include architecture in report.md
I am currently reviewing/fixing a lot of packages for Darwin. It would be handy if bit and architecture would be mentioned by default in the report.md. 

Example:
```
Result of nixpkgs-review run on darwin x84_64
```

```
Result of nixpkgs-review run on NixOS i386
```

This is similar to #129

SuperSandro2000:
Closes #139

I am not sure if we should add 64/32 bit or something else. Open for suggestions.

### DIFF
--- a/nixpkgs_review/report.py
+++ b/nixpkgs_review/report.py
@@ -109,7 +109,10 @@ class Report:
         cmd = "nixpkgs-review"
         if pr is not None:
             cmd += f" pr {pr}"
-        msg = f"Result of `{cmd}` [1](https://github.com/Mic92/nixpkgs-review)\n"
+
+        system = subprocess.check_output(["nix-instantiate", "--eval", "--json", "--expr", "builtins.currentSystem"], text=True).strip('"')
+
+        msg = f"Result of `{cmd}` run on {system} [1](https://github.com/Mic92/nixpkgs-review)\n"
 
         msg += html_pkgs_section(self.broken, "marked as broken and skipped")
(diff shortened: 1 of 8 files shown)


nixpkgs_review/report.py
```diff
@@ -1,4 +1,5 @@
 import os
+import platform
```
Mic92, line 2:
```suggestion
```

nixpkgs_review/report.py
```diff
@@ -109,7 +109,10 @@ def markdown(self, pr: Optional[int]) -> str:
         cmd = "nixpkgs-review"
         if pr is not None:
             cmd += f" pr {pr}"
-        msg = f"Result of `{cmd}` [1](https://github.com/Mic92/nixpkgs-review)\n"
+
+        system = subprocess.check_output(["nix-instantiate", "--eval", "--json", "--expr", "builtins.currentSystem"], text=True).strip('"')
+
+        msg = f"Result of `{cmd}` [1](https://github.com/Mic92/nixpkgs-review) run on {system}\n"
```
SuperSandro2000, line 115:
```suggestion
        msg = f"Result of `{cmd}` run on {system} [1](https://github.com/Mic92/nixpkgs-review)\n"
```

We should probably move this to the end.

SuperSandro2000:
Thanks for the test fixes.

PR #160, opened 2021-01-02
Fix stackoverflow

SuperSandro2000:
Before this if parsing very large amounts of xml I got:

```
error: stack overflow (possible infinite recursion)
```

This happened when reviewing merge PRs which trigger no merge conflict which can happen quite often when reviewing a lot of PRs.

### DIFF
--- a/nixpkgs_review/review.py
+++ b/nixpkgs_review/review.py
@@ -281,6 +281,8 @@ def parse_packages_xml(stdout: IO[bytes]) -> List[Package]:
                 description = value
             elif name == "position":
                 position = value
+        elif event == "end":
+            elem.clear()
     return packages
 
 


nixpkgs_review/review.py
```diff
@@ -281,6 +281,8 @@ def parse_packages_xml(stdout: IO[bytes]) -> List[Package]:
                 description = value
             elif name == "position":
                 position = value
+        elif event == "end":
+            elem.clear()
```
Mic92, line 285: Can you explain how this fix the stackoverflow?

SuperSandro2000: The idea was that it frees resources and worked yesterday. But today I got the same error and running ``ulimit -s 16384`` fixed it. Not sure what goes on here but this may not be the right solution to it.

Mic92: Yeah. Python objects should be created on the heap not the stack. You would get the same effect by starting manual garbage collection.

PR #165, opened 2021-02-02
Override --ff-only option which might be globally configured

turboMaCk:
This fixes the problem with "Not possible to fast-forward..."
in environments where git is configured with `ff = only`

## To reproduce

configure git with `ff only`:

```
$ git config merge.ff only
```

And try to run nix-review in cases where fast forward is not an option.

## Workaround

If you have this in global configuration at the moment the easiest way is to reconfigure nixpkgs repository to override this:

```
$ git config merge.ff false
```

This patch should make it so that merge nixpkgs-review is doing is independent of a configuration.

### DIFF
--- a/nixpkgs_review/review.py
+++ b/nixpkgs_review/review.py
@@ -123,7 +123,7 @@ class Review:
         return str(self.builddir.worktree_dir)
 
     def git_merge(self, commit: str) -> None:
-        sh(["git", "merge", "--no-commit", commit], cwd=self.worktree_dir())
+        sh(["git", "merge", "--no-commit", "--no-ff", commit], cwd=self.worktree_dir())
 
     def apply_unstaged(self, staged: bool = False) -> None:
         args = ["git", "--no-pager", "diff"]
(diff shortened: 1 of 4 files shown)


nixpkgs_review/review.py
```diff
@@ -123,7 +123,7 @@ def worktree_dir(self) -> str:
         return str(self.builddir.worktree_dir)
 
     def git_merge(self, commit: str) -> None:
-        sh(["git", "merge", "--no-commit", commit], cwd=self.worktree_dir())
+        sh(["git", "merge", "--no-commit", "--no-ff", commit], cwd=self.worktree_dir())
```
Mic92, line 126: `--no-ff` or `--ff-only`?

turboMaCk: we want to set it to `--no-ff` because `--ff-only` might not work in some cases.

Mic92: Ok. I made the tests consistent with the actual code.

turboMaCk: Maybe to clarify more this is how I understand these options work in git.

- by default git automatically chooses between fast forward (`--ff`) or merge commit.
- when `--ff-only` git will fail if fast-forward is not possible (this is the setting I use because I like to know that it isn't if it isn't)
- when `--no-ff` git will not attempt to do fast-forward and just creates a merge commit even when fast-forward would be possible. This is likely the fastest and least likely to fail but I never really tried to measure it. If it's faster it will be hardly noticeable anyway.

I think the options are `--ff` and `--no-ff`. I chose the `--no-ff` as it seems more predictable to me.

[documentation](https://www.git-scm.com/docs/merge-options#Documentation/merge-options.txt---ff)

turboMaCk: > Ok. I made the tests consistent with the actual code.

thanks and sorry for that. I've completely missed the inconsistency there. :(

PR #172, opened 2021-02-18
Fix logs for failed builds

SuperSandro2000:
It turns out we need to check both the drv and the path to get all logs when remote building. Only checking the path does not return any failed build logs.

### DIFF
--- a/nixpkgs_review/report.py
+++ b/nixpkgs_review/report.py
@@ -63,18 +63,30 @@ def write_error_logs(attrs: List[Attr], directory: Path) -> None:
                 symlink_source.unlink()
             symlink_source.symlink_to(attr.path)
 
-        if attr.path is not None:
+        if attr.drv_path:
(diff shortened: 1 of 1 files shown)


nixpkgs_review/report.py
```diff
@@ -63,15 +63,16 @@ def write_error_logs(attrs: List[Attr], directory: Path) -> None:
                 symlink_source.unlink()
             symlink_source.symlink_to(attr.path)
 
-        if attr.path is not None:
+        if attr.drv_path or attr.path is not None:
+            path = attr.drv_path or attr.path or ""
```
Mic92, line 67: If path is empty we should not run nix log.

SuperSandro2000: I just left this because mypy was complaing that path is Optional[String]...

SuperSandro2000: Also this approach wont work as both are always set...

Mic92:
fixed in https://github.com/Mic92/nixpkgs-review/pull/174

PR #177, opened 2021-03-02
Fix doCheck

SuperSandro2000:


default.nix
```diff
@@ -17,7 +17,7 @@ python3.pkgs.buildPythonApplication rec {
   ];
 
   # does not work in sandbox
-  dontCheck = false;
+  doCheck = false;
```
Mic92, line 20: Looks like this flag is also causing checkInputs like pytest to be no longer provided, which breaks the test right now.

Mic92: Also we still define a checkPhase in default.nix. I think this one should be executed still.

SuperSandro2000: Then we should remove it because dontCheck does not exist.

Mic92: ok.

SuperSandro2000: force pushed

PR #179, opened 2021-03-04
Unset NIXPKGS_CONFIG to respect meta.broken in review shells

SuperSandro2000:
As discussed in IRC and the meetup. This change respects meta.broken in review-shells.

### DIFF
--- a/nixpkgs_review/review.py
+++ b/nixpkgs_review/review.py
@@ -228,6 +228,7 @@ class Review:
         pr: Optional[int] = None,
         post_result: Optional[bool] = False,
     ) -> None:
+        os.environ.pop("NIXPKGS_CONFIG", None)
         os.environ["NIX_PATH"] = path.as_posix()
         if pr:
             os.environ["PR"] = str(pr)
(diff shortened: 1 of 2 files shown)


nixpkgs_review/review.py
```diff
@@ -228,6 +228,7 @@ def start_review(
         pr: Optional[int] = None,
         post_result: Optional[bool] = False,
     ) -> None:
+        del os.environ["NIXPKGS_CONFIG"]
```
Mic92, line 231: For multiple reviews at the same time this might crash nixpkgs-review if tries to remove a non-existing key:

```suggestion
        os.environ.pop("NIXPKGS_CONFIG", None)
```

SuperSandro2000:
Not sure whats wrong with mypy:

```
## run mypy
nixpkgs_review/__init__.py: error:(B Duplicate module named 'nixpkgs_review' (also at './build/lib/nixpkgs_review/__init__.py')(B
nixpkgs_review/__init__.py: note:(B Are you missing an __init__.py? Alternatively, consider using --exclude to avoid checking one of them.
Found 1 error in 1 file (errors prevented further checking)(B
builder for '/nix/store/mvbn3m3142zgdhl9s31j2z2xbrf1y6gy-nixpkgs-review.drv' failed with exit code 2
```

PR #186, opened 2021-03-25
Display some build output

SuperSandro2000:
to see whats happening

Please merge after the next release.


veprbl:
Does this imply that nix'es default mode is not good?

SuperSandro2000:
The default creates a one high line which is not very useful at all.

Mic92:
@SuperSandro2000 told me, this does not work well in practice.

Issue #171, opened 2021-02-17
Upstream @rmcgibbo fork


SuperSandro2000:
Upstream the changes in https://github.com/rmcgibbo/nixpkgs-review/commits/r-rmcgibbo especially the (feature to merge commits)[https://github.com/rmcgibbo/nixpkgs-review/commit/df28e262d8eaa63d8e4e7166bde7d7be092b9912].

features:
- combine commits https://github.com/rmcgibbo/nixpkgs-review/commit/df28e262d8eaa63d8e4e7166bde7d7be092b9912
- download tarballs for faster checkout in VMs
- ignore packages if they contain a ignored dependency
- and more!

/cc @rmcgibbo

rmcgibbo:
Thanks, @SuperSandro2000.

I just started hacking on this, so I was trying to add the features I wanted for my bot as quickly as possible without really taking the time to file PRs and tests and all that good stuff.

So, some clean up will definitely be required.

In my fork, I've also added hooks at various points to have nixpkgs-review call into external binaries, and I was using those hooks to implement some of the features for my bot:
 -  (a) I want to block the upload of the comment to github under various conditions (e.g. if I can determine that the PR was closed, the builder OOMd, ofborg already built the PR, etc), so I created a pre-github-comment hook
 - (b) I want to run nixpkgs-hammering, so I also do that in the pre-github-comment hook
 - (c) I'm also working on a feature to filter the set of drvs that are built -- kind of like `--skip-packages` but more dynamic -- by running a dry_run build and then solving a (precedence-constrained knapsack problem](https://link.springer.com/article/10.1023%2FA%3A1004649425222) to determine a subset of the build graph that I think I can finish building within a reasonable amount of time. This feature isn't depolyed yet, but my hope is to be able to have r-rmcgibbo build a subset of PRs to staging even if it can't build every drv that changed. Anyways, this is why I introduced the `NIXPKGS_REVIEW_PRE_BUILD_FILTER` hook.
 - (d) Figuring out how long each individual drv took to build is quite difficult. The best way I was able to figure out requires `stat`-ing the file in `/nix/var/log/nix/drvs` where the build log is stored, which is extremely distasteful. And worse, you need to use the new statx system call that only works on some filesystems because the "regular" stat doesn't show the file birth time, and for that you need new code because there's no interface to statx in the python stdlib... Ugh. It's a bit complicated.
 
[...]

PR #187, opened 2021-03-28
Upstream features from r-rmcgibbo bot

rmcgibbo:
I've re-based my fork onto the current top-of-tree, and tried to split my (otherwise disorganized) commits into a logical set of features. Each commit should be separate enough to review independently.

The commits are:
1. Add `--post-logs` feature to post the build logs of failed packages to gist.github.com and to provide a link to the log in the github commit. In order to stay within usage restrictions of gist.github.com, this truncates the logs to 1MB if they're extremely long, and also strips out ANSI color codes. It also takes some care to try to capture stderr from the `nix build` process that might be relevant. For example, hash mismatches are not reported in the log accessible via `nix log`, and so you need to take other measures to capture them and associate them with the log.
2. Add `--prefer-edit` feature. If running nixpkgs-review from multiple systems with the same github account (e.g. aarch64-linux and x86_64-linux), this feature causes the second upload of the github comment to edit the first comment, to avoid an extra ping to each github user subscribed on the thread.
3. pre-build-filter: this feature adds a hook point where `nixpkgs-review` will call and external binary to filter the package set. this hook is used by the r-rmcgibbo bot to implement a dynamic form the the skip-packages feature that tries to skip packages such that the estimated build time falls within a cutoff. This is sufficiently complicated (it involves approximately solving a combinatorial optimization problem called the precedence constrained knapsack problem) that it makes sense as a hook to an external program.
4. A minor commit to reduce the amount of output from git that is sent to stderr, which is useful for keeping the logs clear.

cc #171 @SuperSandro2000

I will ping this PR once my implementation of the pre-build-filter is available. I'm cleaning that up next.


nixpkgs_review/nix.py
```diff
@@ -38,6 +42,61 @@ def was_build(self) -> bool:
...
+        if os.path.exists(full):
+            return full
+        return None
+
+    def build_time(self) -> Optional[timedelta]:
+        log_path = self.log_path()
+        if log_path is None:
+            return None
+
+        proc = subprocess.run(
```
rmcgibbo, line 96: This is ugly, but unfortunately I'm not aware of a better way to get the build time associated with each drv.

1. The idea to use stat on the log file is from: https://github.com/NixOS/nix/issues/1710#issuecomment-470898243
2. The reason for calling the coreutils `stat` binary rather than using the python os.stat function is that cpython does not call the new statx(2) syscall on linux, which is the only way to get the birth time of the file. Note that is not available on all filesystems either. I have a python interface to statx(2) [here](https://github.com/rmcgibbo/statx), but I figured it was easier to use coreutils stat(1) since my statx wrapper is not in nixpkgs and there's been little movement on https://bugs.python.org/issue39533

Mic92: Has this been tested on macOS?

Mic92: You can also use `ctypes` to access statx from libc instead. But than it would be also not portable to macOS since this syscall is os-specific.

default.nix
```diff
@@ -5,6 +5,14 @@ python3.pkgs.buildPythonApplication rec {
   name = "nixpkgs-review";
   src = ./.;
   buildInputs = [ makeWrapper ];
+
+  propagatedBuildInputs = [
+    python3.pkgs.humanize
+    # humanize fails to declare its dependency on septools correctly
+    # https://github.com/NixOS/nixpkgs/pull/113060
```
SuperSandro2000, line 12: That PR is merged.

nixpkgs_review/github.py
```diff
@@ -1,23 +1,35 @@
...
+import urllib.error
 import urllib.parse
 import urllib.request
 from collections import defaultdict
 from typing import Any, DefaultDict, Dict, Optional, Set
 
+import backoff
+
+OWNER = "NixOS"
+REPO = "nixpkgs"
```
SuperSandro2000, line 13: This should use https://github.com/Mic92/nixpkgs-review/blob/12505d5348401b6c842cbe05701774fec9182cee/nixpkgs_review/cli/__init__.py#L185.

nixpkgs_review/github.py
```diff
@@ -1,23 +1,35 @@
+import sys
 import json
+import os
+import urllib.error
 import urllib.parse
 import urllib.request
```
SuperSandro2000, line 6:
```suggestion
from urllib import error, parse, request
```

nixpkgs_review/github.py
```diff
@@ -1,23 +1,35 @@
...
 def pr_url(pr: int) -> str:
-    return f"https://github.com/NixOS/nixpkgs/pull/{pr}"
+    return f"https://github.com/{OWNER}/{REPO}/pull/{pr}"
 
 
 class GithubClient:
     def __init__(self, api_token: Optional[str]) -> None:
         self.api_token = api_token
 
+    @backoff.on_exception(backoff.expo, urllib.error.HTTPError, max_time=60)
```
SuperSandro2000, line 24: Does this mean it waits 60 seconds per request or 60 seconds in total? Might be a bit high.

nixpkgs_review/github.py
```diff
@@ -26,37 +38,67 @@ def _request(
...
+            print(f"Response: {e.read().decode('utf-8')}", file=sys.stderr)
+            raise
+
         return json.loads(resp.read())
 
     def get(self, path: str) -> Any:
         return self._request(path, "GET")
 
-    def post(self, path: str, data: Dict[str, str]) -> Any:
+    def post(self, path: str, data: Dict[str, Any]) -> Any:
```
SuperSandro2000, line 57: Can we be more specific than Any?

nixpkgs_review/github.py
```diff
@@ -26,37 +38,67 @@ def _request(
...
         return self._request(path, "POST", data)
 
     def put(self, path: str) -> Any:
         return self._request(path, "PUT")
 
+    def patch(self, path: str, data: Dict[str, Any]) -> Any:
```
SuperSandro2000, line 63: Can we be more specific than Any?

nixpkgs_review/nix.py
```diff
@@ -38,6 +50,61 @@ def was_build(self) -> bool:
...
+        if self.build_err_msg is not None:
+            value = "\n".join([value, self.build_err_msg])
+
+        return value
+
+    def log_path(self) -> Optional[str]:
+        if self.drv_path is None:
+            return None
+        base = os.path.basename(self.drv_path)
+        prefix = "/nix/var/log/nix/drvs/"
```
SuperSandro2000, line 85: This might now always be true.

nixpkgs_review/nix.py
```diff
@@ -38,6 +50,61 @@ def was_build(self) -> bool:
...
+        full = os.path.join(prefix, base[:2], base[2:] + ".bz2")
```
SuperSandro2000, line 86: logs do not need to be compressed.

nixpkgs_review/nix.py
SuperSandro2000: The darwin equivalent is ``/usr/bin/stat -f "%m" $file``

```
/usr/bin/stat -f "%m" /nix/var/log/nix/drvs/00/074sx96jwrvrj1ypxmzsj2g5k2fsvg-ocaml4.08.1-jackline-unstable-2020-09-03.drv.bz2
1604335341
```

but we can use the stat from coreutils

```
stat --format "%W %Y" /nix/var/log/nix/drvs/00/074sx96jwrvrj1ypxmzsj2g5k2fsvg-ocaml4.08.1-jackline-unstable-2020-09-03.drv.bz2
1604335336 1604335341
```

nixpkgs_review/nix.py
```diff
@@ -89,42 +159,68 @@ def _nix_eval_filter(json: Dict[str, Any]) -> List[Attr]:
...
+
+    #
+    # Split the evaluation into chunks of 4096 attrs at a time.
+    # This helps limit the memory usage, which can be a problem.
+    #
+    start_time = time.time()
+    eval_data = {}
+    attrlist = sorted(attrs)
+    MAX_ATTRS_AT_ONCE = 4096
+    for i in range(0, len(attrs), MAX_ATTRS_AT_ONCE):
```
SuperSandro2000, line 210:
```suggestion
    MAX_ATTRS_AT_ONCE = 4096

    for i in range(0, len(attrs), MAX_ATTRS_AT_ONCE):
```

nixpkgs_review/nix.py
```diff
@@ -89,42 +159,68 @@ def _nix_eval_filter(json: Dict[str, Any]) -> List[Attr]:
...
+            return json.loads(nix_eval.stdout)
+        finally:
+            attr_json.close()
+            if delete:
+                os.unlink(attr_json.name)
+
+    #
+    # Split the evaluation into chunks of 4096 attrs at a time.
+    # This helps limit the memory usage, which can be a problem.
+    #
```
SuperSandro2000, line 205: Does this take longer if you don't mind the RAM usage?

nixpkgs_review/nix.py
```diff
@@ -172,9 +269,79 @@ def nix_build(
...
+    except subprocess.CalledProcessError as e:
+        stderr = e.stderr
+
+    # Remove boring 'copying path' lines from stderr
+    stderr = "\n".join(
+        line
+        for line in stderr.splitlines()
+        if not (
+            line.startswith("copying path '")
+            and line.endswith("from 'https://cache.nixos.org'...")
```
SuperSandro2000, line 283: This does not work for cachix or nix-serve which are used by Mic92's build farm.

nixpkgs_review/nix.py
```diff
@@ -172,9 +269,79 @@ def nix_build(
...
+        )
+    )
+
+    has_failed_dependencies = []
+    has_timeout = {}
+    for line in stderr.splitlines():
+        if "dependencies couldn't be built" in line:
+            has_failed_dependencies.append(
+                next(item for item in line.split() if "/nix/store" in item)
```
SuperSandro2000, line 292: The /nix/store can have some prefix IIRC.

nixpkgs_review/nix.py
```diff
@@ -172,9 +269,79 @@ def nix_build(
...
+                .lstrip("'")
+                .rstrip(":'")
+            )
+        if "timed out after" in line:
+            drv = (
+                next(item for item in line.split() if "/nix/store" in item)
+                .lstrip("'")
+                .rstrip("'")
```
SuperSandro2000, line 300:
```suggestion
                .strip("'")
```

nixpkgs_review/nix.py
```diff
@@ -172,9 +269,79 @@ def nix_build(
...
+            attr.timed_out = True
+
+    for drv_path in has_failed_dependencies:
+        if drv_path in drv_path_to_attr:
+            attr = drv_path_to_attr[drv_path]
+            attr.build_err_msg = stderr
+
+            # Without inspecting the build graph, let's just guess
+            # that if something failed to build and there were ANY timeouts
+            # that it's probably the case that this guy's failure to build
```
SuperSandro2000, line 318:
```suggestion
            # that it's probably the case that this failure to build
```

nixpkgs_review/nix.py
```diff
@@ -172,9 +269,79 @@ def nix_build(
...
+            if len(has_timeout) > 0:
+                attr.timed_out = True
+
+    return attrs
+
+
+def pre_build_filter(attrs: List[Attr]) -> List[Attr]:
+    for cmd in (
+        cmd
+        for cmd in os.environ.get("NIXPKGS_REVIEW_PRE_BUILD_FILTER", "").split(":")
```
SuperSandro2000, line 330: That ENV needs documentation.

nixpkgs_review/report.py
```diff
@@ -117,12 +138,34 @@ def write(self, directory: Path, pr: Optional[int]) -> None:
...
 
-        write_error_logs(self.attrs, directory)
+        write_result_links(self.attrs, directory)
+        write_error_logs(self.failed, directory)
 
     def succeeded(self) -> bool:
         """Whether the report is considered a success or a failure"""
         return len(self.failed) == 0
 
+    def upload_build_logs(self, github_client: GithubClient, pr: Optional[int]) -> None:
```
SuperSandro2000, line 148: We could also implement the fancy logs with colors here https://github.com/SuperSandro2000/nixpkgs-review-checks/blob/d07c4164b14a4fe989acb25eb1428dd795384e26/nixpkgs-review-checks#L323.

nixpkgs_review/report.py
```diff
@@ -147,17 +193,22 @@ def markdown(self, pr: Optional[int]) -> str:
         if pr is not None:
             cmd += f" pr {pr}"
 
-        msg = f"Result of `{cmd}` run on {self.system} [1](https://github.com/Mic92/nixpkgs-review)\n"
+        shortcommit = f" at {self.pr_rev[:8]}" if self.pr_rev else ""
+        link = "[1](https://github.com/Mic92/nixpkgs-review)"
+        msg = f"Result of `{cmd}`{shortcommit} run on {self.system} {link}\n"
 
-        msg += html_pkgs_section(self.broken, "marked as broken and skipped")
+        msg += html_pkgs_section(self.broken, "marked as broken and skipped", show=10)
```
SuperSandro2000, line 200: I think we shouldn't limit the contents of lists in details by default.

setup.cfg
```diff
@@ -7,14 +7,15 @@ ignore = E501,E741,W503
...
-ignore = E501,E741,W503
+ignore = E501,E741,W503,E203
 exclude = .git,__pycache__,docs/source/conf.py,old,build,dist
 
 [mypy]
 warn_redundant_casts = true
 disallow_untyped_calls = true
 disallow_untyped_defs = true
 no_implicit_optional = true
+ignore_missing_imports = true
```
SuperSandro2000, line 18: Why do we want to ignore this?

SuperSandro2000:
Thanks for doing this! ❤️

nixpkgs_review/nix.py
rmcgibbo: I linked above to the python wrapper for statx that I have. It uses ctypes. https://github.com/rmcgibbo/statx/blob/master/statx.py

nixpkgs_review/github.py
Mic92: It's exponential backoff. It will increase waiting time between each try up to one minute.

default.nix
rmcgibbo: Is that sufficient? I personally call this default.nix with nixpkgs from 20.09.

nixpkgs_review/github.py
rmcgibbo: Not really. You can post any JSON data to the API. The gist API requires some nested dicts, which is the reason for the change.

nixpkgs_review/github.py
rmcgibbo: I don't think so.

nixpkgs_review/nix.py
rmcgibbo: Not really, since the chunks are so big (how often do you have 4096 attrs anyways?).

nixpkgs_review/report.py
rmcgibbo: I think in this case, it makes sense to put this feature (uploading the logs) into `nixpkgs-review` rather than having it as a custom extension. Doing it as a `gist` allows us to re-use the same GITHUB_TOKEN infrastructure, which makes a lot of sense.

nixpkgs_review/report.py
rmcgibbo: When I run nixpkgs-review against staging in `r-rmcgibbo`, there are sometimes up to 10,000 attrs "skipped" by the pre-build-hook filter, so I think some amount of limiting is necessary. Perhaps it can apply only to the "skipped" category rather than the "broken" category?

setup.cfg
rmcgibbo: The reason this is required is that there are no mypy headers for the `@backoff` decorator.

default.nix
SuperSandro2000: We can backport this is required.

nixpkgs_review/nix.py
SuperSandro2000: When rebuilding staging that is quite common.

nixpkgs_review/report.py
SuperSandro2000: I wanted to say that we can also throw the colored logs into nixpkgs-review

nixpkgs_review/report.py
SuperSandro2000: Currently we don't list skipped packages at all so we did not have this problem. I think applying it to skipped packages is fine I think.

setup.cfg
SuperSandro2000: Can we ignore this inline then globally?

nixpkgs_review/github.py
rmcgibbo: Unfortuately that doesn't work, since they're used in the code as, for example, `urllib.error` rather than as `error`.

nixpkgs_review/nix.py
rmcgibbo: I don't think it really effects the time very much at all.

nixpkgs_review/nix.py
rmcgibbo: I don't actually know how to determine this location robustly. It could be in a different location if you're running with a sufficiently esoteric `/etc/nix/nix.conf`, and it's not even guaranteed that nix's config file with be in `/etc/nix/nix/conf`.

nixpkgs_review/nix.py
rmcgibbo: I can remove all "copying path '" from the logs. It's just for convenience though.

nixpkgs_review/nix.py
rmcgibbo: Yeah, I'm not sure what the right way to determine what the prefix actually is, if it's not /nix/store. Look at the start of `$(which nix)`? :P

Mic92: You can evaluate `builtins.storeDir`

setup.cfg
Mic92: It is also possible to limit this per package instead for all imports. See example below in this file.

nixpkgs_review/nix.py
rmcgibbo: Marked with a "TODO" item

nixpkgs_review/report.py
rmcgibbo: Maybe we can do that in a second PR?

default.nix
rmcgibbo: Would you be able to do that? I've never done a backport.

nixpkgs_review/nix.py
rmcgibbo: This is a potentially area of discussion, @Mic92.

Despite being a small amount of code, for me this is the most important feature. The idea is that I want to run some code to manipulate the set of `attrs` that the rest of the code deals with, to set some as `skipped=True` based on a dynamic calculation that I can control. Basically a "plugin".

I can think of two possible ways to achieve this feature. One is as written, where `nixpkgs-review` connects with these external plugins by environment variable and subprocesses them. A different way to do it that you might prefer would be to use `import.metadata` `entry_points`, which is kind of a python plugin feature. We could agree on the name of the entry point, and then this `pre_build_filter` function would find the plugins by iterating over the matching entry points rather than looking through an environment variable. The big difference here would be that the plugin would execute inside the same process as the main python interpreter.

I am open to either mechanism.

default.nix
SuperSandro2000: https://github.com/NixOS/nixpkgs/pull/118153

nixpkgs_review/github.py
SuperSandro2000: Wouldn't it be enough to just import urllib then?

nixpkgs_review/nix.py
```diff
@@ -204,3 +377,38 @@ def write_shell_expression(filename: Path, attrs: List[str], system: str) -> Non
...
+            [ -/]*  # Intermediate bytes
+            [@-~]   # Final byte
+        )
+    """,
+        re.VERBOSE,
+    )
+    return ansi_escape.sub("", s)
+
+
+def _store_dir() -> str:
```
SuperSandro2000, line 402: Can well tell python to cache this? The nix store path will never change while running.

nixpkgs_review/nix.py
SuperSandro2000: Can we use ``_store_dir()`` here?

nixpkgs_review/github.py
```diff
@@ -83,10 +126,20 @@ def get_borg_eval_gist(self, pr: Dict[str, Any]) -> Optional[Dict[str, Set[str]]
...
+        if raw_gist_url is not None:
+            for line in urllib.request.urlopen(raw_gist_url):
+                if line == b"":
+                    break
+                system, attribute = line.decode("utf-8").split()
+                packages_per_system[system].add(attribute)
+            return packages_per_system
         return None
+
+    def upload_gist(self, name: str, content: str, description: str) -> Dict[str, Any]:
```
SuperSandro2000, line 140: Can we add an option to use a different token for posting gists? I have them on my bot account but would like to post as me.

nixpkgs_review/github.py
rmcgibbo: No — it doesn’t automatically import the sub modules. I tested that.

nixpkgs_review/nix.py
rmcgibbo: It's only called once anyways, but yes I can add a `@functools.lru_cache`.

nixpkgs_review/github.py
rmcgibbo: Would you want to control this by providing an environment variable like `GITHUB_GIST_TOKEN` that would have a higher precedence than the other mechanisms for setting the token?

SuperSandro2000: Thats how I do it right now. Would be great if you could add this :)

nixpkgs_review/nix.py
Mic92: I will have a look this evening into this PR. Thanks for leading the review so far @SuperSandro2000 . Due to diverse requirements and languages used in the community a language-independent solution might be preferred.

Mic92: @ryantm is this something that nixpkgs-update could benefit from as well? If not what is missing?

default.nix
```diff
@@ -5,6 +5,14 @@ python3.pkgs.buildPythonApplication rec {
   name = "nixpkgs-review";
   src = ./.;
   buildInputs = [ makeWrapper ];
+
+  propagatedBuildInputs = [
+    python3.pkgs.humanize
```
Mic92, line 10: Eventually I want get rid of humanize/backport again, but I can also do this later on master.
Right now we are unsetting PYTHONPATH to avoid injecting dependencies into nix-shell.

nixpkgs_review/github.py
Mic92: How about adding build logs within the comment instead of a gist? https://github.com/NixOS/nixpkgs/pull/125133#issuecomment-851825296

rmcgibbo: That seems strictly worse to me. At least in my browser (firefox), there's no dedicated HTML page for files embedded within the comment, so glancing at each log requires downloading the file and navigating to the downloads folder and opening it up.

Mic92: Fair enough.

default.nix
rmcgibbo: Do you prefer if I vendor the `naturaldelta` function from `humanize` into this PR, or we could also do something based on wrappers so that we get `python3.pkgs.humanize` on sys.path even though external settings of PYTHONPATH are ignored?

SuperSandro2000: Vendoring the code is probably less hacky. Maybe we can also rewrite the function if it is less complex.

nixpkgs_review/nix.py
rmcgibbo: I added some documentation to the README.md

happysalada:
@rmcgibbo thank you for this! ❤️

SuperSandro2000:
@rmcgibbo Can you take a look at the tests failures? I think it is unable to parse remotes that are local files.

PR #202, opened 2021-05-29
buildenv.py: don't set git author and committer name

veprbl:
This prevents from creating misattributed commits when trying to do any git work on nixpkgs in the resulting shell.


SuperSandro2000:
I rely on this ENV here https://github.com/SuperSandro2000/nixpkgs-review-checks/blob/master/bashrc#L122

veprbl:
git merge is called with --no-commit, so no commits are created

PR #216, opened 2021-08-16
cli/__init__.py: support GitHub's auth token format in oauth_token

zakame:
Use a non-capturing group to include `ghp_` prefixes in reading
`oauth_token` from `hub` config, while keeping compatibility with older
tokens (that the user should be refreshing soon.)

cf https://github.blog/2021-04-05-behind-githubs-new-authentication-token-formats/

### DIFF
--- a/nixpkgs_review/cli/__init__.py
+++ b/nixpkgs_review/cli/__init__.py
@@ -121,7 +121,7 @@ def read_github_token() -> Optional[str]:
         try:
             with open(path) as f:
                 for line in f:
-                    token_match = re.match(r"\s*oauth_token:\s+([a-f0-9]+)", line)
+                    token_match = re.match(r"\s*oauth_token:\s+((?:ghp_)?[a-f0-9]+)", line)
                     if token_match:
                         return token_match.group(1)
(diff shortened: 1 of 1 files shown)


nixpkgs_review/cli/__init__.py
```diff
@@ -121,7 +121,7 @@ def read_github_token() -> Optional[str]:
         try:
             with open(path) as f:
                 for line in f:
-                    token_match = re.match(r"\s*oauth_token:\s+([a-f0-9]+)", line)
+                    token_match = re.match(r"\s*oauth_token:\s+((?:gh[pousr]_)?[a-f0-9]+)", line)
```
SuperSandro2000, line 124: I think we only support ghp tokens and I think we should limit the filter to them.

zakame: Fine with me, that'd make this change much simpler!  Will update soon unless anyone else chimes in :muscle:

zakame:
Thanks @SuperSandro2000! 🎉

Mic92:
I think we should make a release for this.

PR #249, opened 2022-01-15
Fix error log generation

0xbe7a:
Currently, only the log of the first derivative is saved and the function returns afterwards. Instead it should break and continue with the next derivation


Mic92 (approved):
Thanks!

nixpkgs_review/report.py
```diff
@@ -79,7 +79,7 @@ def write_error_logs(attrs: List[Attr], directory: Path) -> None:
                     stdout=f,
                 )
                 if nix_log.returncode == 0:
-                    return
+                    break
```
Mic92, line 82: His return and break here not equivalent? There is nothing after this loop?

PR #254, opened 2022-02-01
Eval fixes

jonringer:
Keep running into these issues when reviewing staging PRs

### DIFF
--- a/nixpkgs_review/nix.py
+++ b/nixpkgs_review/nix.py
@@ -52,7 +52,7 @@ def nix_shell(
         raise RuntimeError("nix-shell not found in PATH")
 
     shell = cache_directory.joinpath("shell.nix")
-    write_shell_expression(shell, attrs, system)
+    write_shell_expression(shell, attrs, system, True)
     if sandbox:
         args = _nix_shell_sandbox(nix_shell, shell)
     else:
@@ -233,7 +233,7 @@ def nix_build(
         return attrs
 
     build = cache_directory.joinpath("build.nix")
-    write_shell_expression(build, filtered, system)
+    write_shell_expression(build, filtered, system, allow_aliases)
 
     command = [
         "nix",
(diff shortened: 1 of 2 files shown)


jonringer:
Testing this on some PRs. Takes a while for `nix-env -qa` to run.

jonringer:
cc @Mic92 looks ready to go, was able to eval https://github.com/NixOS/nixpkgs/pull/142593 by adding `--disable-aliases`

SuperSandro2000 (approved):
ofborg sometimes puts some aliases in the eval results. I am not sure why but it seems to always be gnome.*

jonringer:
> ofborg sometimes puts some aliases in the eval results. I am not sure why but it seems to always be gnome.*

unless you're reviewing a PR made against master, it will force the eval to include aliases

```
 --disable-aliases     Disable aliases while evaluating locally. They are automatically disabled when reviewing
                        NixOS/nixpkgs PRs against master.
```

SuperSandro2000:
Yeah, we put that in because PRs from stable could still contain aliases and it is not easy to remove them there.

Maybe those lines need to be updated https://github.com/Mic92/nixpkgs-review/blob/ab4a35df2c885f29e336aa3c3e30734283989b80/nixpkgs_review/cli/pr.py#L41-L44

The intention was, that only aliases are forbidden when reviewing PRs merged against master.

jonringer:
> Yeah, we put that in because PRs from stable could still contain aliases and it is not easy to remove them there.

There's an ofborg check to prevent that situation. There should be no alias usage within nixpkgs.

example PR: https://github.com/NixOS/nixpkgs/pull/156318

```
 ofborg-eval-package-list-no-aliases — nix-env -qa --json --file . --arg config { allowAliases = false; }
```

jonringer:
> The intention was, that only aliases are forbidden when reviewing PRs merged against master.

#255

SuperSandro2000:
I merged #255 and pushed 1ad19620ef898df35f80b9ee03d18bf80dfbe003 to master.

PR #255, opened 2022-02-01
Disable aliases by default

jonringer:
Ofborg does a check with `allowAliases = false;`. So it's safe to assume that our evaluation shouldn't have to worry about them either.

```
 ofborg-eval-package-list-no-aliases — nix-env -qa --json --file . --arg config { allowAliases = false; }
```

related: https://github.com/Mic92/nixpkgs-review/pull/254

### DIFF
--- a/nixpkgs_review/cli/__init__.py
+++ b/nixpkgs_review/cli/__init__.py
@@ -151,10 +151,10 @@ def read_github_token() -> Optional[str]:
 def common_flags() -> List[CommonFlag]:
     return [
         CommonFlag(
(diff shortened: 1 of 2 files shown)


jonringer:
This would be a cli breaking change. Not sure if you want to do the semver bump or not

SuperSandro2000:
> This would be a cli breaking change. Not sure if you want to do the semver bump or not

We probably release it as 2.7.0 in a bit.

SuperSandro2000 (approved):

PR #256, opened 2022-02-01
Remove now aliases nixUnstable

SuperSandro2000:
@Mic92 or do you have a better idea?

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -15,7 +15,7 @@ python3.pkgs.buildPythonApplication rec {
 
     # needed for interactive unittests
     python3.pkgs.pytest
-    nix_2_4
+    pkgs.nixVersions.nix_2_4 or nix_2_4
     git
   ];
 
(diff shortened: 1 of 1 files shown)


default.nix
```diff
@@ -15,7 +15,7 @@ python3.pkgs.buildPythonApplication rec {
 
     # needed for interactive unittests
     python3.pkgs.pytest
-    nix_2_4
+    nixVersions.nix_2_4
```
Mic92, line 18: Does this work in 21.11 as well?

SuperSandro2000: I don't think so but I also don't know how to make it work in both versions.

jtojnar: You could use `pkgs.nixVersions.nix_2_4 or nix_2_4`.

SuperSandro2000: That does not work because nixVersions is also undefined.

jtojnar: the `or` syntax should work even for nested attribute sets.

SuperSandro2000: Now it works. I tried that before and I am sure that it didn't work.

PR #273, opened 2022-05-28
Disable external git diff tools when patching

bcc32:
The GIT_EXTERNAL_DIFF environment variable can be configured to use a
custom diff tool for interactive diff commands, but it may not produce
an output compatible with `git apply`, so `--no-ext-diff` should be
used to ensure that the diff output is usable for this purpose.


Mic92:
Thanks!

PR #286, opened 2022-12-20
make --system work better with the shell

figsoda:
without this pr (on nixos):

```console
$ nixpkgs-review rev HEAD --system aarch64-darwin
[...]

$ ls
bash: /nix/store/npcq66nqv2wpzp3ma5s63vafnnivphxb-coreutils-9.1/bin/ls: cannot execute binary file: Exec format error
```

this pr changes so the shell is built with the local system, and the snippet works

### DIFF
--- a/nixpkgs_review/nix.py
+++ b/nixpkgs_review/nix.py
@@ -275,7 +275,7 @@ let
   paths = [
 """
         )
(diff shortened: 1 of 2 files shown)


SuperSandro2000 (commented):
didn't test but LGTM

PR #293, opened 2022-12-24
disallow url literals by default

figsoda:
this is the default behavior for ofborg and hydra: https://github.com/NixOS/ofborg/pull/628

also adds `--allow-url-literals`

### DIFF
--- a/nixpkgs_review/cli/__init__.py
+++ b/nixpkgs_review/cli/__init__.py
@@ -152,22 +152,11 @@ def read_github_token() -> Optional[str]:
 def common_flags() -> List[CommonFlag]:
     return [
         CommonFlag(
-            "--allow-aliases",
-            dest="allow_aliases",
-            action="store_true",
-            help="Allow aliases while evaluating locally",
-        ),
-        CommonFlag(
-            "--allow-ifd",
-            "--allow-import-from-derivation",
-            dest="allow_ifd",
-            action="store_true",
-            help="Allow import-from-derivation",
-        ),
-        CommonFlag(
-            "--allow-url-literals",
-            action="store_true",
-            help="Allow url literals",
+            "--allow",
+            action="append",
+            default=[],
+            choices=["aliases", "ifd", "url-literals"],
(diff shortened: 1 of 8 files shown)


SuperSandro2000:
What speaks against settings this globally in nix.conf?

figsoda:
That would require the user to do it, and users don't have to follow nixpkgs conventions for all of their nix projects

SuperSandro2000:
true but ofborg is also checking it now. Not sure if we really need it but it is a nice addition.

figsoda:
url literals are probably just supported for the sake of backward compatibility, I'm not sure if there is anything else nixpkgs-review can do to change that

nixpkgs_review/cli/__init__.py
```diff
@@ -164,6 +164,11 @@ def common_flags() -> List[CommonFlag]:
             action="store_true",
             help="Allow import-from-derivation",
         ),
+        CommonFlag(
+            "--allow-url-literals",
```
Mic92, line 168: I think I see a pattern here. I don't want to so many `--allow-$X` flags here. Instead a I would prefer a `--allow=ifd,aliases,url-literals` flag.

Mic92: Since this is mostly interactive usage I think we can safely drop existing flags without causing too much damage.

Mic92: If we have than some datastructure that can store all those flags than we don't need to pass down as many parameters to other functions.

figsoda: Sounds like a good idea, should I keep the --allow-aliases flag to show a deprecation message?

figsoda: Instead of comma separated, I opted for it beeing passed multiple times: `--allow ifd --allow aliases` since it is easier to implement with `choices`

Mic92: I think we can just drop the flag.

Mic92:
bors merge

PR #303, opened 2023-02-11
feature: add `--nom` flag to enable nix-output-monitor

### LINKED ISSUE #302: Option to use nix-output-monitor
For large builds nix-output-monitor can give a really nice overview of the current progress. Here is a screenshot from the [project page](https://github.com/maralorn/nix-output-monitor):

nom-build

This is much more informative than the normal `nix build` progress bar.

Please add an option to use `nom-build`/`nom build` instead of `nix-build`/`nix build`.

gador:
`nix-output-monitor` provides a nice real time build graph, which gives more information about what builds where, which is especially nice for larger rebuilds.

fixes #302

### DIFF
--- a/README.md
+++ b/README.md
@@ -153,8 +153,7 @@ nix-shell> nixpkgs-review comments
 ```
 
 `nixpkgs-review` will by default use [nix-output-monitor](https://git.maralorn.de/nix-output-monitor/about/) if found in `$PATH`.
-If you have a custom path to `nom`, this can be passed with the `--nom-path` flag.
-Also, if you have `nom` installed but don't want to use it, you can pass an empty string to `--nom-path` to use `nix build` instead of `nom build`.
+If you have `nom` installed but don't want to use it, you can pass `nix` to `--nix-flavor` to use `nix build` instead of `nom build`.
 
 ## Using nixpkgs-review in scripts or other programs
 
--- a/README.md
+++ b/README.md
@@ -152,7 +152,7 @@ nix-shell> nixpkgs-review post-result
 nix-shell> nixpkgs-review comments
 ```
 
-For larger rebuilds, it can come in handy to see which packages still need to be build. This can be achieved with the `--nom` flag. This enables the `nix-output-monitor` (https://git.maralorn.de/nix-output-monitor/about/) which informs about remote builders, down- and uploads and displays a build graph.
+For larger rebuilds, it can come in handy to see which packages still need to be build. This can be achieved with the `--nom` flag. This enables the [nix-output-monitor](https://git.maralorn.de/nix-output-monitor/about/) which informs about remote builders, down- and uploads and displays a build graph.
 
 To use this feature, make sure to install `nix-output-monitor` first.
 
--- a/README.md
+++ b/README.md
@@ -152,8 +152,8 @@ nix-shell> nixpkgs-review post-result
 nix-shell> nixpkgs-review comments
 ```
 
-`nixpkgs-review` will by default use [nix-output-monitor](https://git.maralorn.de/nix-output-monitor/about/) if found in `$PATH`.
-If you have `nom` installed but don't want to use it, you can pass `nix` to `--nix-flavor` to use `nix build` instead of `nom build`.
+`nixpkgs-review` will by default use [nix-output-monitor](https://github.com/maralorn/nix-output-monitor) if found in `$PATH`.
+If you have `nom` installed but don't want to use it, you can pass `nix` to `--build-graph` to use `nix build` instead of `nom build`.
 
 ## Using nixpkgs-review in scripts or other programs
 
--- a/README.md
+++ b/README.md
@@ -152,6 +152,10 @@ nix-shell> nixpkgs-review post-result
 nix-shell> nixpkgs-review comments
 ```
 
+For larger rebuilds, it can come in handy to see which packages still need to be build. This can be achieved with the `--nom` flag. This enables the `nix-output-monitor` (https://git.maralorn.de/nix-output-monitor/about/) which informs about remote builders, down- and uploads and displays a build graph.
+
+To use this feature, make sure to install `nix-output-monitor` first.
+
 ## Using nixpkgs-review in scripts or other programs
 
 After building, `nixpkgs-review` will normally start a `nix-shell` with the
--- a/README.md
+++ b/README.md
@@ -152,9 +152,9 @@ nix-shell> nixpkgs-review post-result
 nix-shell> nixpkgs-review comments
 ```
 
-For larger rebuilds, it can come in handy to see which packages still need to be build. This can be achieved with the `--nom` flag. This enables the [nix-output-monitor](https://git.maralorn.de/nix-output-monitor/about/) which informs about remote builders, down- and uploads and displays a build graph.
-
-To use this feature, make sure to install `nix-output-monitor` first.
+`nixpkgs-review` will by default use [nix-output-monitor](https://git.maralorn.de/nix-output-monitor/about/) if found in `$PATH`.
+If you have a custom path to `nom`, this can be passed with the `--nom-path` flag.
+Also, if you have `nom` installed but don't want to use it, you can pass an empty string to `--nom-path` to use `nix build` instead of `nom build`.
 
 ## Using nixpkgs-review in scripts or other programs
 
--- a/nixpkgs_review/cli/__init__.py
+++ b/nixpkgs_review/cli/__init__.py
@@ -228,7 +228,7 @@ def common_flags() -> List[CommonFlag]:
             type=str,
             default=nix_nom_tool(),
             choices=["nix", "nom"],
-            help='Build graph to use. Use either "nom" or "nix". Will default to "nom" if available',
+            help='Build graph to print. Use either "nom" or "nix". Will default to "nom" if available',
         ),
     ]
 
--- a/nixpkgs_review/cli/__init__.py
+++ b/nixpkgs_review/cli/__init__.py
@@ -1,6 +1,7 @@
 import argparse
 import os
 import re
+import shutil
 import sys
 from pathlib import Path
 from typing import Any, List, Optional, Pattern, cast
@@ -223,6 +224,11 @@ def common_flags() -> List[CommonFlag]:
             default=read_github_token(),
             help="Github access token (optional if request limit exceeds)",
         ),
+        CommonFlag(
+            "--nom",
+            action="store_true",
+            help="Use nix-output-monitor (nom) instead of nix. Provides a nice realtime build graph",
+        ),
     ]
 
 
@@ -285,6 +291,13 @@ def check_common_flags(args: argparse.Namespace) -> bool:
     elif args.no_shell:
         print("--no-shell and --run are mutually exclusive", file=sys.stderr)
         return False
+    if args.nom:
+        if not shutil.which("nom"):
+            print(
+                "Cannot find nix-output-monitor. Please install it first if you want to use it",
+                file=sys.stderr,
+            )
+            return False
     return True
 
 
--- a/nixpkgs_review/cli/__init__.py
+++ b/nixpkgs_review/cli/__init__.py
@@ -224,11 +224,11 @@ def common_flags() -> List[CommonFlag]:
             help="Github access token (optional if request limit exceeds)",
         ),
         CommonFlag(
-            "--nix-flavor",
+            "--build-graph",
             type=str,
             default=nix_nom_tool(),
             choices=["nix", "nom"],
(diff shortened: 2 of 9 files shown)


README.md
```diff
@@ -152,6 +152,10 @@ nix-shell> nixpkgs-review post-result
 nix-shell> nixpkgs-review comments
 ```
 
+For larger rebuilds, it can come in handy to see which packages still need to be build. This can be achieved with the `--nom` flag. This enables the `nix-output-monitor` (https://git.maralorn.de/nix-output-monitor/about/) which informs about remote builders, down- and uploads and displays a build graph.
```
figsoda, line 155:
```suggestion
For larger rebuilds, it can come in handy to see which packages still need to be build. This can be achieved with the `--nom` flag. This enables the [nix-output-monitor](https://git.maralorn.de/nix-output-monitor/about/) which informs about remote builders, down- and uploads and displays a build graph.
```

README.md
```diff
@@ -152,6 +152,10 @@ nix-shell> nixpkgs-review post-result
...
+
+To use this feature, make sure to install `nix-output-monitor` first.
```
figsoda, line 157: We might want to have an option in the package like bubblewrap so it can be more easily tested in CI

nixpkgs_review/nix.py
```diff
@@ -246,17 +247,30 @@ def nix_build(
...
-        "--extra-experimental-features",
-        "nix-command" if allow.url_literals else "nix-command no-url-literals",
-        "build",
-        "--no-link",
-        "--keep-going",
-        "--allow-import-from-derivation"
-        if allow.ifd
-        else "--no-allow-import-from-derivation",
-    ]
+    if nom:
```
figsoda, line 250: this whole command does not need to be duplicated, `nom` should have a compatible CLI and the only thing we need to change is `"nix"`

figsoda (changes requested):
There are a few more uses of nix tools that should be changed, you can find all the instances by running `rg '"nix(-\w+)?"'`

nixpkgs_review/nix.py
gador: Thats what I did first. It does not work, because `nom` expects all build arguments after `build`.  This is why the order of arguments is slightly changed.

gador:
> There are a few more uses of nix tools that should be changed, you can find all the instances by running `rg '"nix(-\w+)?"'`

I disagree. `nom` only makes sense when `build` is used. E.g. https://github.com/Mic92/nixpkgs-review/blob/d7a07874b4651a0945583b5b73faaf912d6b4947/nixpkgs_review/utils.py#L48-L59 doesn't make any sense to use `nom`

nixpkgs_review/nix.py
figsoda: `nix` should be able to handle build arguments after `build`, we can change the order of the `nix` invocations to be the same as `nom`'s

figsoda:
turns out most things that pop up from rg command wouldn't work with nom, but the `nix-shell` in `nix.py` could be changed to `nom-shell`

gador:
I'm not too familiar with the code that uses `nix-shell`. But it was my understanding, that the shell gets executed after everything is build. So, in my understanding, changing this to `nom-shell` doesn't yield any improvements, right?

figsoda:
`nom-shell` doesn't do much more than doing a `mkShell` and `buildEnv`, but I still think it would be nice to have a consistent interface for commands that could be replaced by nom

nixpkgs_review/cli/__init__.py
```diff
@@ -223,6 +224,11 @@ def common_flags() -> List[CommonFlag]:
             default=read_github_token(),
             help="Github access token (optional if request limit exceeds)",
         ),
+        CommonFlag(
+            "--nom",
+            action="store_true",
```
Mic92, line 229: Should this be just enabled by default if `nom` is in $PATH?

Mic92: I think this could become a string option pointing to the nom executable that defaults to the nom discovered in `$PATH`. If someone wants to disable nom instead they would than pass an empty string to this option.

Mic92:
I am ok with @figsoda reviewing this as long as nom stays optional (maybe automatically discovered if present).

Mic92 (approved):
I am ok with this feature this as long as nom stays optional (maybe automatically discovered if present). If I don't come back to this just merge it if it looks good to you @figsoda.

figsoda (approved):
lgtm

having an option like bubblewrap is the only thing can be implemented in a separate pr

gador:
@figsoda Shall we change it to autodiscover nom and enable it by default? Or leave it as is?

figsoda:
that sounds like a cool idea, I'll leave the decision to @Mic92

gador:
I took the liberty to adjust the PR as per @Mic92 suggestion. We can revert the commit, if we decide to not use it.

Otherwise it now uses `nom` by default, if it is found. Also I added some test cases.

nixpkgs_review/utils.py
```diff
@@ -9,7 +10,7 @@
 ROOT = Path(os.path.dirname(os.path.realpath(__file__)))
 
 
-def color_text(code: int, file: IO[Any] = sys.stdout) -> Callable[[str], None]:
```
gador, line 12: If you are wondering why I changed this:

I needed access to `sys.stdout` for the test cases. For some reason `pytest` cannot access the standard output correctly. (See https://github.com/pytest-dev/pytest/issues/5997 or https://github.com/pytest-dev/pytest/issues/6676) .
The comment https://github.com/pytest-dev/pytest/issues/6676#issuecomment-582393940 mentioned using `None` as output. This still outputs everything correctly to the terminal, but allows the tests to access stdout as well.

As per [the docs](https://docs.python.org/3/library/functions.html#print) this defaults to its original behavior anyway:
>The file argument must be an object with a write(string) method; if it is not present or None, [sys.stdout](https://docs.python.org/3/library/sys.html#sys.stdout) will be used.

figsoda:
instead of specifying the path, what if we have something like `--nix-flavor`, empty string will default to nom and fall back to nix, `nom` will force nom, and `nix` will force nix, this way we can also allow `nom-shell` to be used

gador:
I like the `nix-flavor` better. See the new commit

nixpkgs_review/nix.py
```diff
@@ -45,12 +45,13 @@ def nix_shell(
     attrs: List[str],
     cache_directory: Path,
     system: str,
+    nix_flavor: str,
     run: Optional[str] = None,
     sandbox: bool = False,
 ) -> None:
-    nix_shell = shutil.which("nix-shell")
+    nix_shell = shutil.which(nix_flavor)
```
figsoda, line 52: this is using `nix` and `nom` instead of `nix-shell` and `nom-shell`

nixpkgs_review/cli/pr.py
```diff
@@ -66,6 +66,7 @@ def pr_command(args: argparse.Namespace) -> str:
                     allow=allow,
                     checkout=checkout_option,
                     sandbox=args.sandbox,
+                    nix_flavor=args.nix_flavor,
```
figsoda, line 69: you can resolve nix_flavor into paths to `nix` and `nix-shell` here

nixpkgs_review/utils.py
```diff
@@ -62,3 +63,11 @@ def current_system() -> str:
         text=True,
     )
     return system.stdout
+
+
+def nix_nom_tool() -> str:
```
figsoda, line 68: this logic can probably go into the review comment above

> you can resolve nix_flavor into paths to nix and nix-shell here

README.md
```diff
@@ -152,6 +152,9 @@ nix-shell> nixpkgs-review post-result
 nix-shell> nixpkgs-review comments
 ```
 
+`nixpkgs-review` will by default use [nix-output-monitor](https://git.maralorn.de/nix-output-monitor/about/) if found in `$PATH`.
```
SuperSandro2000, line 155:
```suggestion
`nixpkgs-review` will by default use [nix-output-monitor](https://github.com/maralorn/nix-output-monitor) if found in `$PATH`.
```
we should link to the github repo where people can create issues and PRs

nixpkgs_review/cli/__init__.py
```diff
@@ -223,6 +223,13 @@ def common_flags() -> List[CommonFlag]:
             default=read_github_token(),
             help="Github access token (optional if request limit exceeds)",
         ),
+        CommonFlag(
+            "--nix-flavor",
```
SuperSandro2000, line 227: nom is not a nix flavor but a wrapper around nix itself.

figsoda: I struggled to find a good name for this, do you have any suggestions?

nixpkgs_review/cli/pr.py
gador: I would like to not resolve it to a path. This would keep things simple, as only `nix` or `nom` is passed and should work as well, since it is in `PATH`.

nixpkgs_review/utils.py
gador: see above

nixpkgs_review/cli/__init__.py
gador: I think `--build-graph` is a good description

figsoda (approved):
Not entirely sure about `--build-graph` either, but I don't have a better name, so lgtm

gador:
shall I squash the commits and rebase?

figsoda:
you can rebase and force push if you want, or this can also be squash merged

gador:
Ok, then you can squash merge if it seems all alright to you :+1:

nixpkgs_review/cli/__init__.py
```diff
@@ -223,6 +223,13 @@ def common_flags() -> List[CommonFlag]:
             default=read_github_token(),
             help="Github access token (optional if request limit exceeds)",
         ),
+        CommonFlag(
+            "--build-graph",
+            type=str,
+            default=nix_nom_tool(),
+            choices=["nix", "nom"],
+            help='Build graph to use. Use either "nom" or "nix". Will default to "nom" if available',
```
SuperSandro2000, line 231: For me --build-graph sounds more like a boolean option which when passed to nixpkgs-review would trigger to use nom. The help is not that fitting for me.

gador: If this isn't helpful, I'd appreciate a suggestion. 
For me, the initial implementation using `--nom` was a boolean. `--build-graph` (to me) doesn't sound like a boolean. I'd rather expect `--use-build-graph` or similar to be a boolean. But I agree, it isn't optimal, yet.
If you have a suggestion how to phrase it more distinctly, please let me know.

Mic92: I think it's fine, if you use `--help` you get this:

```
  --build-graph {nix,nom}
                        Build graph to use. Use either "nom" or "nix". Will default to "nom" if available
```

Mic92, line 231:
```suggestion
            help='Build graph to print. Use either "nom" or "nix". Will default to "nom" if available',
```

nixpkgs_review/cli/__init__.py
```diff
@@ -223,6 +223,13 @@ def common_flags() -> List[CommonFlag]:
             default=read_github_token(),
             help="Github access token (optional if request limit exceeds)",
         ),
+        CommonFlag(
+            "--build-graph",
```
Mic92, line 227: How about this name?

```suggestion
            "--build-program",
```

nixpkgs_review/cli/__init__.py
gador: I like it. It's clearer that it is basically just prints differently

nixpkgs_review/cli/__init__.py
gador: I like the `graph` more, since that is the main feature why we'd use `nom`.  `--build-program` sounds like it will build the derivation with something different (which isn't true. It is just a pipe from `nix` to `nom`)

nixpkgs_review/utils.py
```diff
@@ -62,3 +63,11 @@ def current_system() -> str:
...
+    "Return `nom` if found in $PATH"
+    if not shutil.which("nom"):
+        return "nix"
+    else:
+        return "nom"
```
Mic92, line 73: nitpick simplification:

```suggestion
    if shutil.which("nom"):
        return "nom"
    else:
        return "nix"
```

davidak (approved):
This is a very helpful feature as it gives an overview of what is being built.

Works great. Not reviewed the code in detail. No obvious issue found.

hmenke (approved):

Mic92:
bors merge

PR #312, opened 2023-02-25
Add --print-result flag

### LINKED ISSUE #311: Printing the result comment to stdout instead of posting it
There is `nixpkgs-review post-result`, but it requires a GitHub access token, and I’m not really comfortable giving `nixpkgs-review` access to one. If `nixpkgs-review` could print the comment that it would post, I would happily paste it into the comment box in my browser and post it manually.

figsoda:
closes #311


Mic92:
bors merge

PR #315, opened 2023-02-27
add --extra-nixpkgs-config

### LINKED ISSUE #314: Allow passing extra arguments to `import <nixpkgs>`
Use-case: running `nixpkgs-review` with `config.cudaSupport = true` and custom `config.cudaCapabilities`.
Currently, `nixpkgs-review` only tracks affected cuda-enabled packages if they expose a `xxxWithCuda` attribute in nixpkgs.
Not all packages do that, and it's debatable whether they should.

I think a satisfiable solution would be e.g. being able to run `nixpkgs-review pr XXXXXX --extra-nixpkgs-config '{ config.cudaSupport = true; }'`. I imagine `nixpkgs-review` could explicitly state any extra options it has received in the report that it publishes as a github comment, since these options could (and do) affect the eval

figsoda:
closes #314


README.md
```diff
@@ -163,6 +163,13 @@ nix-shell> nixpkgs-review comments
 `nixpkgs-review` will by default use [nix-output-monitor](https://github.com/maralorn/nix-output-monitor) if found in `$PATH`.
 If you have `nom` installed but don't want to use it, you can pass `nix` to `--build-graph` to use `nix build` instead of `nom build`.
 
+Some pull requests may require configuration for nixpkgs to test out. You can
+use the `--extra-nixpkgs-config` flag to supply extra configuration for nixpkgs.
+
+```console
+$ nixpkgs-review pr 37242 --extra-nixpkgs-config '{ config.cudaSupport = true; }'
```
SuperSandro2000, line 170: We should omit the ``{}`` or enforce/check them otherwise things will break.

SomeoneSerge:
Is there anything to be done with `list_packages`? https://github.com/Mic92/nixpkgs-review/blob/6ffaffadeaafaf054fa5e76180789e387734fa6a/nixpkgs_review/review.py#L340

README.md
figsoda: I don't think we should omit `{}` because the argument would not be valid nix, unless we want something more abstract like `--extra-nixpkgs-config config.cudaSupport true`, the trailing semicolon can be unintuitive to users

maybe we can evaluate it first without nixpkgs so the error messages will be better? enforcing `{}` can have false positives if things builtin functions are used

figsoda:
I'm not sure. There are a few things we can do

- use `--extra-nixpkgs-config` and do nothing to `nix-env` (what we are doing right now)
- use `--extra-nixpkgs-arg` (and `--extra-nixpkgs-argstr`) instead, similar to `nix-env`'s `--arg` and `--argstr`
- use `--extra-nixpkgs-config` and pass it to `nix-env` (with `-E`?)

SomeoneSerge:
IIUC, we need either of the last two options, otherwise `review.differences` will miss the affected attributes

README.md
SuperSandro2000: It should be fine to enforce people to supply something like ``{ ... }``. They should not code in this anyway.

Mic92:
I will skip reviewing this feature but I think we should have some way of passing nixpkgs configuration into nixpkgs-review. Just merge if you come to a consensus @SuperSandro2000 @figsoda

README.md
figsoda: > use `--extra-nixpkgs-arg` (and `--extra-nixpkgs-argstr`) instead, similar to `nix-env`'s `--arg` and `--argstr`

What do you think about this option? @SuperSandro2000

nixpkgs_review/nix/evalAttrs.nix
```diff
@@ -1,8 +1,14 @@
-{ allowAliases ? false, attr-json }:
+{ allowAliases ? false, attr-json, extra-nixpkgs-config }:
 
 with builtins;
 let
-  pkgs = import <nixpkgs> { config = { checkMeta = true; allowUnfree = true; inherit allowAliases; }; };
+  pkgs = import <nixpkgs> ({
+    config = {
+      checkMeta = true;
+      allowUnfree = true; inherit allowAliases;
```
SomeoneSerge, line 8: If `extra-nixpkgs-config` contains `config` as an attribute, we're going to loose `allowUnfree`. We either need a recursive merge or we stick to `--nixpkgs-arg[str]` instead

SomeoneSerge: Here's another bit, I think: https://github.com/Mic92/nixpkgs-review/blob/6ffaffadeaafaf054fa5e76180789e387734fa6a/nixpkgs_review/buildenv.py#L38-L43

SomeoneSerge:
Here's a preview of the current status: https://gist.github.com/SomeoneSerge/745681c4d577fac2b2e71b6349cc2503 with this quick solution to `differences()`: https://github.com/SomeoneSerge/nixpkgs-review/commit/9c3c9fe82f27cee5a9ff340068296b943a494e89

At a glance, I think the list of packages looks good

Mic92:
When this PR is merged I will make a new release.

figsoda:
One issue I have currently is that nix-env doesn't do recursive updates. Would any config change the outputs of nix-env? I'm thinking we should just ignore nix-env and implement recursive updates for the regular eval

SomeoneSerge:
I think I'm observing some false-positive "updates": https://gist.github.com/SomeoneSerge/6cc00b41964e43f725fc12046778532d

SomeoneSerge:
Is it expect behaviour that `nixpkgs-review pr` doesn't always print the "X packages updated" part? E.g. here it didn't mention `nix-env` and just went ahead to `nix build`: https://gist.github.com/5c5ceccb1add211d551a9813ef8be960

figsoda:
It is expected I believe: https://github.com/Mic92/nixpkgs-review/commit/e704d6a5b694f79881247f4f39aee12d63bd6575

nixpkgs_review/buildenv.py
```diff
@@ -20,11 +19,32 @@ def find_nixpkgs_root() -> Optional[str]:
...
+  ## TODO: also build packages marked as insecure
+  # allowInsecurePredicate = x: true;
+}} // {extra_nixpkgs_config}
+"""
+            )
+        )
+        self.nixpkgs_config.flush()
 
-    def __enter__(self) -> None:
+    def __enter__(self) -> str:
```
SomeoneSerge, line 47: Maybe make it `Path`? It's understandable here, but gets confusing by the time it reaches `review_local_revision(..., nixpkgs_config: str, ...)`

SomeoneSerge (approved):
The current version works for me (example: https://github.com/NixOS/nixpkgs/pull/218265#issuecomment-1451721482) and I'm happy with the UX. I find it confusing that for some PRs the tool doesn't show in advance which packages it's going to build, but I'm not even sure if it's related to this change. The rest is in Sandro's scope, I suppose:)

README.md
```diff
@@ -163,6 +163,13 @@ nix-shell> nixpkgs-review comments
 `nixpkgs-review` will by default use [nix-output-monitor](https://github.com/maralorn/nix-output-monitor) if found in `$PATH`.
 If you have `nom` installed but don't want to use it, you can pass `nix` to `--build-graph` to use `nix build` instead of `nom build`.
 
+Some pull requests may require configuration for nixpkgs to test out. You can
+use the `--extra-nixpkgs-config` flag to supply extra configuration for nixpkgs.
+
+```console
+$ nixpkgs-review pr 37242 --extra-nixpkgs-config '{ cudaSupport = true; }'
```
Mic92, line 170: Since we are now adding more option that influence how nix evaluation is done, can we add this to `report.md` as well? This would help the community a lot if some posts a build result for reproduction.

SomeoneSerge: It does go into the report:
image

figsoda:
@SuperSandro2000

nixpkgs_review/buildenv.py
```diff
@@ -20,11 +20,32 @@ def find_nixpkgs_root() -> Optional[str]:
...
 
-@dataclass
 class Buildenv:
-    allow_aliases: bool
+    def __init__(self, allow_aliases: bool, extra_nixpkgs_config: str) -> None:
+        if not (
+            extra_nixpkgs_config.startswith("{") and extra_nixpkgs_config.endswith("}")
+        ):
+            raise RuntimeError(
+                "--extra-nixpkgs-config should start with `{` and end with `}`"
```
SuperSandro2000, line 29:
```suggestion
                "--extra-nixpkgs-config must start with `{` and end with `}`"
```
should should not abort the execution

figsoda:
bors r+

I messaged sandro privately. He is currently prioritizing on other things and is not against merging this

nviets:
Can I use this argument to cap resources with max-jobs and cores? By default nixpkgs-review tries to build with everything.

SuperSandro2000:
You want to set those settings in your nix.conf.

PR #339, opened 2023-06-19
print help mesesage when run with no flags

figsoda:


nixpkgs_review/cli/__init__.py
```diff
@@ -289,6 +289,10 @@ def parse_args(command: str, args: List[str]) -> argparse.Namespace:
         for flag in common:
             parser.add_argument(*flag.args, **flag.kwargs)
 
+    if args == []:
+        main_parser.print_help()
+        exit(0)
```
l0b0, line 294: Shouldn't this be a non-zero exit code? Running without a subcommand is an error.

figsoda: I thought it might be a bit confusing when the command fails but the message doesn't say why it fails

emilazy: `git(1)` and `brew(1)` both print commands and exit nonzero when run without arguments, I think that's the convention we should follow.

PR #349, opened 2023-08-11
don't rely on `NIX_PATH`

### LINKED ISSUE #343: `os.environ["NIX_PATH"] = self.nixpkgs_path()` silently get ignored when `nix-path` is set in `nix.conf`
```bash
$ grep nix-path /etc/nix/nix.conf
nix-path =
$ echo $NIX_PATH

$ nix eval --expr 'builtins.nixPath' --impure
[ ]
$ export NIX_PATH=foo=bar
$ nix eval --expr 'builtins.nixPath' --impure
[ ]
$ nix eval --expr 'builtins.nixPath' --impure --option nix-path foo=baz
[ { path = "baz"; prefix = "foo"; } ]
```

This may or may not be considered a bug of nix. But it breaks `nixpkgs-review` when `nix.channel.enable = false;` is set.

Reference: https://github.com/NixOS/nixpkgs/pull/242098 https://github.com/NixOS/nixpkgs/blob/b56ed968c4840efdc871f7408997b1d9c07bf76a/nixos/modules/config/nix-channel.nix#L98

figsoda:
Nix does not respect `NIX_PATH` when the `nix-path` setting in nix.conf
is set, which breaks the `nix-store --verify-path` due to `<nixpkgs>`
diverging from the temporary nixpkgs created by nixpkgs-review.

fixes #343

### DIFF
--- a/nixpkgs_review/builddir.py
+++ b/nixpkgs_review/builddir.py
@@ -58,7 +58,10 @@ class Builddir:
 
         self.worktree_dir = self.path.joinpath("nixpkgs")
         self.worktree_dir.mkdir()
-        nix_path = ["nixpkgs={self.worktree_dir}", "nixpkgs-overlays={self.overlay.path}"]
+        nix_path = [
+            f"nixpkgs={self.worktree_dir}",
+            f"nixpkgs-overlays={self.overlay.path}",
+        ]
         # we don't actually use this, but its handy for users who want to try out things with the current nixpkgs version.
         os.environ["NIX_PATH"] = ":".join(nix_path)
         self.nix_path = " ".join(nix_path)
--- a/nixpkgs_review/builddir.py
+++ b/nixpkgs_review/builddir.py
@@ -66,10 +66,6 @@ class Builddir:
         os.environ["NIX_PATH"] = ":".join(nix_path)
         self.nix_path = " ".join(nix_path)
 
-        self.nix_path = (
-            f"nixpkgs={self.worktree_dir} nixpkgs-overlays={self.overlay.path}"
-        )
-
     def __enter__(self) -> "Builddir":
         return self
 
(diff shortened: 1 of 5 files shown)


nixpkgs_review/builddir.py
```diff
@@ -63,7 +63,7 @@ def __init__(self, name: str) -> None:
         os.environ["NIX_PATH"] = self.nixpkgs_path()
 
     def nixpkgs_path(self) -> str:
-        return f"nixpkgs={self.worktree_dir}:nixpkgs-overlays={self.overlay.path}"
+        return f"nixpkgs={self.worktree_dir} nixpkgs-overlays={self.overlay.path}"
```
Mic92, line 66: Is space an allowed separator in the environment variable as well?

figsoda: reading the source code, I don't think it is

figsoda: I rewrote the patch to make nixpkgs-review not rely on NIX_PATH, hopefully I didn't accidently break something

Mic92: I actually find it quite handy to have a NIX_PATH set. Can we not just keep it, but not rely on it for our own tooling?

nixpkgs_review/builddir.py
```diff
@@ -58,12 +58,10 @@ def __init__(self, name: str) -> None:
 
         self.worktree_dir = self.path.joinpath("nixpkgs")
         self.worktree_dir.mkdir()
```
Mic92, line 60:
```suggestion
        self.worktree_dir.mkdir()
        nix_path = ["nixpkgs={self.worktree_dir}", "nixpkgs-overlays={self.overlay.path}"]
        os.environ["NIX_PATH"] = ":".join(nix_path)
        self.nix_path = " ".join(nix_path)
```

Mic92:
@mergify queue

PR #351, opened 2023-08-17
fix typos

figsoda:
This is technically a breaking change because of the change in the json schema (`non-existant` -> `non-existent`)


Mic92:
@mergify merge

Mic92:
@mergify queue

PR #352, opened 2023-08-25
allow access to the nixpkgs overlay in the sandbox

figsoda:
fixes this warning

```
warning: Nix search path entry '/tmp/tmpXXXXXXXX' does not exist, ignoring
```


figsoda:
@mergify queue

PR #366, opened 2023-10-11
feat: add `--hook` to run commands when done thinking

cafkafk:
This adds a `--hook` flag to the `pr` subcommand.

This is useful for e.g. running `notify-send "done building"`, sending mail, or
any arbitrary thing you might wanna do when `nixpkgs-review` is done bulding.

It's been a long time since I wrote any python, and I have no grasp on this
codebase, so appologies if I made a mess!


Artturin:
test fail

nixpkgs_review/cli/__init__.py
```diff
@@ -56,6 +56,12 @@ def pr_flags(
         choices=["merge", "commit"],
         help=checkout_help,
     )
+    pr_parser.add_argument(
+        "--hook",
+        default="",
```
Mic92, line 61:
```suggestion
```

nixpkgs_review/review.py
```diff
@@ -579,6 +584,7 @@ def review_local_revision(
             build_graph=args.build_graph,
             nixpkgs_config=nixpkgs_config,
             extra_nixpkgs_config=args.extra_nixpkgs_config,
+            hook=args.hook,
```
Mic92, line 587: This Result class also needs this field. You can use `mypy nixpkgs_review` to test it for type errors like this.

cafkafk: When I run it I get:

```
mypy nixpkgs_review
Success: no issues found in 20 source files
```

Not sure what a result class is in this context

nixpkgs_review/cli/__init__.py
cafkafk: fixed!

cafkafk:
I'm not sure I understand python enough to fix this for now, might reopen later

nixpkgs_review/review.py
Artturin: > When I run it I get:
> 
> ```
> mypy nixpkgs_review
> Success: no issues found in 20 source files
> ```
> 
> Not sure what a result class is in this context

Rebase the branch, before https://github.com/Mic92/nixpkgs-review/pull/371 a --strict was needed

PR #371, opened 2023-10-19
add pylint configuration

Artturin:
but not tests yet

### DIFF
--- a/nixpkgs_review/nix.py
+++ b/nixpkgs_review/nix.py
@@ -297,7 +297,7 @@ def nix_build(
 def write_shell_expression(
     filename: Path, attrs: list[str], system: str, nixpkgs_config: Path
 ) -> None:
-    with open(filename, "w+") as f:
+    with open(filename, "w+", encoding="utf-8") as f:
         f.write(
             f"""{{ pkgs ? import ./nixpkgs {{ system = \"{system}\"; config = import {nixpkgs_config}; }} }}:
 with pkgs;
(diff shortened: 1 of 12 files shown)


Artturin:
Will fix lints in a separate PR

figsoda:
Ruff seems to support some pylint lints under `PL`, since we are already using ruff, can we use ruff instead? Adding `"PL"` to `tools.ruff.select` might be sufficient

nixpkgs_review/nix.py
```diff
@@ -297,7 +297,7 @@ def nix_build(
 def write_shell_expression(
     filename: Path, attrs: list[str], system: str, nixpkgs_config: Path
 ) -> None:
-    with open(filename, "w+") as f:
+    with open(filename, "w+", encoding="utf-8") as f:
```
Mic92, line 300: What happens if you don't specify encoding?

Mic92:
Here is a feature comparison: https://github.com/astral-sh/ruff/blob/d31a700f773b97b265109b63bbeb0ef994f63a3b/README.md#how-does-ruff-compare-to-pylint

Mic92:
I guess it doesn't hurt.. checking should be still fast enough

Mic92:
@mergify queue

nixpkgs_review/nix.py
Artturin: https://pylint.pycqa.org/en/latest/user_guide/messages/warning/unspecified-encoding.html

> It is better to specify an encoding when opening documents. Using the system default implicitly can create problems on other operating systems. See https://peps.python.org/pep-0597/

PR #372, opened 2023-10-20
pylint fixes

Artturin:
fix use-dict-literal

https://pylint.pycqa.org/en/latest/user_guide/messages/refactor/use-dict-literal.html


Mic92:
@mergify queue

PR #373, opened 2023-10-20
Make the `review-shell` purer

### LINKED ISSUE #370: summary is hidden behind lots of collisions
I often run nixpkgs-review on vimPlugins update or the lua package update set, at the moment https://github.com/NixOS/nixpkgs/pull/258819 . 

In both cases, the summary is followed by a long list of collisions that makes the summary a pain to find:
```
warning: collision between `/nix/store/82icynaidvzyic2g349cqi4p9a5d4aa2-vimplugin-ai.vim-2023-10-03/doc/tags' and `/nix/store/qc4bdwi6zhk4r1vzbcxrnwrrjnm33a6g-vimplugin-CheckAttach-2019-05-08/doc/tags'
warning: collision between `/nix/store/82icynaidvzyic2g349cqi4p9a5d4aa2-vimplugin-ai.vim-2023-10-03/.gitignore' and `/nix/store/7k2xjb9fd6x0k2myb04759cd6sl0w9pc-vimplugin-ChatGPT.nvim-2023-10-16/.gitignore'
warning: collision between `/nix/store/82icynaidvzyic2g349cqi4p9a5d4aa2-vimplugin-ai.vim-2023-10-03/README.md' and `/nix/store/hsshj9pwkm5di7vha04haas87hfxya8b-vimplugin-BetterLua.vim-2020-08-14/README.md'
warning: collision between `/nix/store/82icynaidvzyic2g349cqi4p9a5d4aa2-vimplugin-ai.vim-2023-10-03/LICENSE.txt' and `/nix/store/kxbsb0yhk9z70dxr73jmyzwm4sfw8aa0-vimplugin-LanguageClient-neovim-0.1.161/LICENSE.txt'
```
I haven't looked at the code but the log shows that nixpkgs-review creates a big environment that contains everything. Would it make sense to ignore collisions in that environment ? why is the env needed ?

Artturin:
The only purpose of this `env` (I assume) is to add the binaries to the `PATH`.

The benefit of making it purer is that unwrapped programs missing dependencies may fail.

As a added benefit this dramatically reduces the collision warnings from `buildEnv`

`warning: collision between '/nix/store/82icynaidvzyic2g349cqi4p9a5d4aa2-vimplugin-ai.vim-2023-10-03/doc/tags' and '/nix/store/qc4bdwi6zhk4r1vzbcxrnwrrjnm33a6g-vimplugin-CheckAttach-2019-05-08/doc/tags'`

Closes https://github.com/Mic92/nixpkgs-review/issues/370

### DIFF
--- a/nixpkgs_review/nix.py
+++ b/nixpkgs_review/nix.py
@@ -300,22 +300,23 @@ def write_shell_expression(
     with open(filename, "w+", encoding="utf-8") as f:
         f.write(
             f"""{{ pkgs ? import ./nixpkgs {{ system = \"{system}\"; config = import {nixpkgs_config}; }} }}:
-with pkgs;
+
 let
-  paths = [
+  paths = with pkgs; [
 """
         )
         f.write("\n".join(f"    {escape_attr(a)}" for a in attrs))
         f.write(
             """
   ];
-  env = buildEnv {
+  hostPkgs = (import ./nixpkgs { });
+  env = hostPkgs.buildEnv {
     name = "env";
     inherit paths;
     pathsToLink = [ "/bin" ];
     ignoreCollisions = true;
   };
-in (import ./nixpkgs { }).mkShell {
+in hostPkgs.mkShell {
   name = "review-shell";
   preferLocalBuild = true;
   allowSubstitutes = false;
--- a/nixpkgs_review/nix.py
+++ b/nixpkgs_review/nix.py
@@ -312,6 +312,7 @@ let
   env = buildEnv {
     name = "env";
     inherit paths;
+    pathsToLink = [ "/bin" ];
     ignoreCollisions = true;
   };
 in (import ./nixpkgs { }).mkShell {
--- a/nixpkgs_review/nix.py
+++ b/nixpkgs_review/nix.py
@@ -320,7 +320,8 @@ in (import ./nixpkgs { }).mkShell {
   preferLocalBuild = true;
   allowSubstitutes = false;
   dontWrapQtApps = true;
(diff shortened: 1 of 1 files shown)


nixpkgs_review/nix.py
Artturin, line 1: @Mic92 suggested that the `  packages = if builtins.length paths > 50 then [ env ] else paths;` should be changed to use the `buildEnv` always

Issues with that

There's no python3.11 in PATH when using `buildEnv` because it doesn't take into account the `propagatedBuildInputs` when reviewing PR 261696 a python module bump

There's no `$PYTHONPATH` or other variables set because no hooks

However, are we sure that those aren't good things? Binaries will fail to launch if they're missing deps(good) and reviewing by importing modules is often not useful because there's already a `pythonImportsCheck` in python packages.

The hooks just bring side effects which shouldn't be relied on outside nix builds, nixpkgs-review should be testing without side effects

Artturin: `"/${python3.sitePackages}"`
could be in `buildEnv` and in the `review-shell`
`PYTHONPATH = "${env}/${python3.sitePackages};`
and `python3` appended to `packages`

Mic92: What would be than our suggested workflow for to users of the tool to test libraries?

Mic92:
@mergify queue

Mic92:
@mergifyio refresh

Mic92:
@mergify queue

Theghost256 (approved):

Mic92:
Superseded by #384, and since #690 the review shell no longer goes through nix at all (bin/ of built outputs is put on PATH directly).

PR #382, opened 2023-12-20
nix: make review-shell.nix code static

Atemu:
Previously, the whole shell.nix was created at runtime using string substitution and other crimes that made it hard to work on the expression itself.

This patch turns the string substitutions into Nix function arguments to an otherwise static Nix expression and only generates the list of attrs to be built dynamically.


Mic92:
@mergify queue

Atemu:
Thanks!

PR #384, opened 2024-01-02
always use the buildEnv

Mic92:
The purpose of this `env` is to add the binaries to the `PATH`.

The benefit of making it purer is that unwrapped programs missing dependencies may fail.

As a added benefit this dramatically reduces the collision warnings from `buildEnv`

based on https://github.com/Mic92/nixpkgs-review/pull/373

### DIFF
--- a/nixpkgs_review/nix/review-shell.nix
+++ b/nixpkgs_review/nix/review-shell.nix
@@ -6,8 +6,6 @@
   # Path to Nix file containing a list of attributes to build
   nixpkgs-path,
   # Path to this review's nixpkgs
-  alt-pkgs ? null,
-  # Alternative package set name (e.g. pkgsCross.aarch64-multiplatform), or null for default
   local-pkgs ? import nixpkgs-path {
     system = local-system;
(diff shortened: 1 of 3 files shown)


Theghost256 (approved):

nixpkgs_review/nix/review-shell.nix
```diff
@@ -3,21 +3,18 @@
...
   allowSubstitutes = false;
   dontWrapQtApps = true;
-  packages = if builtins.length attrs > 50 then [ env ] else attrs;
+
+  packages = pkgs.buildEnv {
+    name = "env";
+    paths = import attrs-path pkgs;
+    pathsToLink = [ "/bin" ];
+    ignoreCollisions = true;
+  };
```
ShamrockLee: IIRC, `packages` need to be a list.

```suggestion
  packages = lib.singleton (
    pkgs.buildEnv {
      name = "env";
      paths = import attrs-path pkgs;
      pathsToLink = [ "/bin" ];
      ignoreCollisions = true;
    }
  );
```

PR #388, opened 2024-02-25
fix case where nom-shell is in not in PATH

Mic92:


Artturin:
https://github.com/Mic92/nixpkgs-review/pull/389 was merged

PR #397, opened 2024-04-10
Add flag to build package `passthru.tests`

### LINKED ISSUE #77: Feature request: Build passthru.tests automatically
It would be helpful to run passthru.tests automatically when they exists like [nixpkgs-update](https://github.com/ryantm/nixpkgs-update/pull/118) and [ofborg](https://github.com/NixOS/ofborg/issues/368) (will) do.

B4dM4n:
This PR adds a `--tests` flag to build `passthru.tests` for all packages, as discussed in #77.

The tests to be build can be filtered with `-p`, `-P` and similar, like all other packages.

```
$ nixpkgs-review pr --tests 303033
...
Link to currently reviewing PR:
https://github.com/NixOS/nixpkgs/pull/303033

1 tests built:
netbird.passthru.tests.netbird

2 packages built:
netbird netbird-ui
...
```

```
$ nixpkgs-review pr --tests --package-regex 'netbird.+' 303033
...
1 tests built:
netbird.passthru.tests.netbird

1 package built:
netbird-ui
...
```

Since they are classified as tests, they can also be enabled individually with `-p`, without specifying `--tests`:
```
$ nixpkgs-review pr -p netbird.passthru.tests.netbird 303033
...
1 tests built:
netbird.passthru.tests.netbird
...
```

Fixes: #77

### DIFF
--- a/nixpkgs_review/review.py
+++ b/nixpkgs_review/review.py
@@ -638,11 +638,8 @@ class Review:
         if not self.shell_options.no_shell:
             shell_config = ShellConfig(
                 cache_directory=path,
-                local_system=self.build_config.local_system,
                 build_graph=self.shell_options.build_graph,
                 nix_path=self.builddir.nix_path,
-                nixpkgs_config=self.build_config.nixpkgs_config,
-                nixpkgs_overlay=self.builddir.overlay.path,
                 run=self.shell_options.run,
                 sandbox=self.shell_options.sandbox,
             )
@@ -1037,7 +1034,6 @@ def build_config_from_args(
     return BuildConfig(
         allow=allow,
         nix_path=nix_path,
-        local_system=current_system(),
         nixpkgs_config=nixpkgs_config,
         num_eval_workers=args.num_eval_workers,
         max_memory_size=args.max_memory_size,
--- a/nixpkgs_review/review.py
+++ b/nixpkgs_review/review.py
@@ -606,6 +606,7 @@ class Review:
         report = Report(
             commit,
             attrs_per_system,
+            self.build_config,
             self.package_filter,
             ReportOptions(
                 extra_nixpkgs_config=self.review_config.extra_nixpkgs_config,
@@ -807,7 +808,7 @@ def _join_packages_for_system(
     changed_attrs: dict[Path, Attr],
     specified_attrs: dict[Path, Attr],
 ) -> dict[Path, Attr]:
-    # ofborg does not include tests and manual evaluation is too expensive
+    # ofborg does not include tests, so don't mark them as nonexistent
     tests = {path for path, attr in specified_attrs.items() if attr.is_test()}
 
     nonexistent = specified_attrs.keys() - changed_attrs.keys() - tests
@@ -1052,6 +1053,8 @@ def build_config_from_args(
     allow: AllowedFeatures,
     nix_path: str,
     nixpkgs_config: Path,
+    *,
+    include_tests: bool,
 ) -> BuildConfig:
     """Create a BuildConfig from parsed CLI arguments."""
     return BuildConfig(
@@ -1060,6 +1063,7 @@ def build_config_from_args(
         nixpkgs_config=nixpkgs_config,
         num_eval_workers=args.num_eval_workers,
         max_memory_size=args.max_memory_size,
+        include_tests=include_tests,
     )
 
 
--- a/nixpkgs_review/review.py
+++ b/nixpkgs_review/review.py
@@ -461,17 +461,22 @@ class Review:
     def build(
         self, packages_per_system: dict[System, set[str]], args: str
     ) -> dict[System, list[Attr]]:
-        packages_per_system = filter_packages_per_system(
+        attrs_per_system = filter_packages_per_system(
             packages_per_system,
             self.package_filter,
             self.build_config,
         )
-        packages_per_system = {
-            system: self.package_filter.additional_packages | packages
-            for system, packages in packages_per_system.items()
+        additional_attrs_per_system = eval_packages_per_system(
+            dict.fromkeys(self.systems, self.package_filter.additional_packages),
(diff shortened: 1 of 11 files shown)


nixpkgs_review/review.py
```diff
@@ -472,9 +489,24 @@ def filter_packages(
...
+    if build_passthru_tests and len(specified_packages) == 0:
+        changed_attrs = package_attrs(
+            changed_packages,
+            system,
+            allow,
+            nix_path,
+            build_passthru_tests=build_passthru_tests,
+        )
+        changed_packages = set(
+            changed_attrs[path].name for path in changed_attrs.keys()
```
Mic92, line 507: Sorry for the late reply. I think because of the potential large memory consumption of nixos tests, we need to evaluate them isolation as they will otherwise make the build potentially go out of memory, i.e. if you try to have a large package set rebuild.
Therefore I would suggest to evaluate those packages ahead of time i.e. using `nix-instantiate` and add them to the final `nix-build`. Of course a better design would be some sort of pipe-lining where already evaluated packages gets build as the rest of the evaluation is still in progress. But this probably requires bigger refactorings and could implemented at a later point similar to nix-fast-build.

Theghost256 (approved):

nixpkgs_review/review.py
lucasew: Maybe at first we could do this way but warn about this possible problem and then address it in a follow up PR

Mic92: I don't think it's a good idea to just merge it like that as going out of memory can also causes the users system to crash in some cases i.e. crashing their wayland compositor.

Eveeifyeve (changes requested):
As indicated in suggestion comment, please fix the memory corruption bugs via mic92 suggestion and additionally solve the conflicts.

LunNova:
Is OOMing really an issue unique to running tests? There are some very memory hungry normal packages you can accidentally hit in a review run and we don't insist on nixpkgs-review having a special job scheduler to handle that well.

As long as running tests is behind an optional flag it's probably fine to not have a great solution for them being memory hungry.

Eveeifyeve:
> Personally, I never had any problems with memory consumption while evaluating packages/tests, because I only run `nixpkgs-review` on machines which can do a local package evaluation via `nix-env -qaP`, which itself currently requires around 22GB of memory. So everything that consumes less just works fine.
> 
> But I get that most people probably only review PR's with ofborg evaluation, so the memory consumption in this case is (usually) much lower.
> 
> When performing local evaluation, I currently observe the following memory consumption:
> 
>     * 22GB per `nix-env -qaP` invocation
> 
>     * 4GB usage by the `nixpkgs-review` python process after the `nix-env -qaP` runs. Not sure what is causing that, maybe the produced XML output is kept in memory?
> 
> 
> Afterwards the numbers heavily depend on the number of changed packages:
> 
>     * `nix eval` potential unbounded memory usage (which could rapidly grow when evaluating tests)
> 
>     * `nix build` potentially also has unbounded memory usage, because it's evaluating the same attributes again
> 
> 
> To make the memory consumption a bit more manageable, I tried to implement local valuation with `nix-eval-jobs` in #473, which comes with its own issues regarding speed/performance, so it's probably not happening soon (`nix-eval-jobs` would need a "lazy eval" mode like `nix-env -qaP` has, to be viable).
> 
> Based on my experience from that PR, I reworked this PR to only use `nix-eval-jobs` to evaluate the changed packages and tests with a configurable memory "limit":
> 
> [master...B4dM4n:nixpkgs-review:build-passthru-tests-v2](https://github.com/Mic92/nixpkgs-review/compare/master...B4dM4n:nixpkgs-review:build-passthru-tests-v2)
> 
> It also removes the potential huge and permanent memory consumption of `nix build`, by passing it the wanted derivations directly, instead of evaluating the attributes again. There are currently no gcroots created for the derivations, but I will add that soon.
> 
> I could also try to rewrite the current PR, to invoke one `nix-instantiate` command per test attribute to limit the memory usage, but this seems rather wasteful.
> 
> Which route would you prefer?

`nix-instantiate` then `nix-build` because you can have large evals and it prepares the build so build doesn't have to do eval then build it can just build.

B4dM4n:
I rebased my v2 branch onto current master and restructured the commits a bit.

The first three commits are performance/resource usage improvements, which could be useful on their own, but are essential to test evaluation.

Summary:

- Use nix-eval-jobs in `multi_system_eval`:
  This provides the basic infrastructure, to limit resource consumption during package evaluation. Multiple systems will now be evaluated in the same nix-eval-jobs invocation.

- Don't evaluate packages in `nix build`:
  Previously, the attribute paths of the packages to be rebuild would be passed to `nix build`, forcing it to evaluate all packages again, right after `multi_system_eval` has already done that. Passing the result of `multi_system_eval` to `nix build`, removes that step and reduces the memory consumptiom of `nix build` quite a lot.

- Reduce redundant package evaluations:
  In many cases, `filter_packages` will already evaluate the packages to be be rebuild and `nix_build` would do it again afterwards. So doing it always in `filter_packages` removes the need for the second evaluation.

The commit introducing the actual `--tests` flag is quite minimal after those preparations.

The preparation commits could also be reviewed in order in separate PR's if desired (they depend only on the previous commits, but are otherwise fully functional themselves)

samestep:
In case it's helpful to anyone, since the first two commits of this PR are now redundant with what's already in `main`, here's the third commit 47f4647579439762e2f2b1618aee3abd9b60c83b rebased onto the current `main` commit 7f919b017be41c9a716bbd7fdb77fa829e96f3fa (disclaimer: the rebasing was done by Claude):

[**`rebased.patch`**](https://github.com/user-attachments/files/29713758/rebased.patch)

Mic92:
Landed in reduced form as #691 on top of #690. Thanks for the groundwork!

PR #399, opened 2024-05-11
print copiable commands

Et7f3:
[shlex.join](https://docs.python.org/3/library/shlex.html#shlex.join) handle adding ' when argument contains space. Before it was printed like this: `--extra-experimental-features nix-command no-url-literals`


Mic92:
@mergify queue

PR #403, opened 2024-06-08
Add checkout strategy to report.

pbsds:
This way, `nixpkgs-review pr ... -c commit` will add `--checkout commit` to the markdown report. This is useful when checking against hydra


Mic92:
@mergify queue

Mic92:
@mergify queue

Issue #415, opened 2024-09-08
nixpkgs-review doesn't differentiate between build and store errors.


CyberShadow:
Hi, I'm a relatively new nixpkgs contributor. nixpkgs-review is advertised very visibly in the PR template, so I've been trying to use it for my own PRs.

However, one issue I have is that when nixpkgs-review reports that something failed to build, it doesn't provide any other information.

Today, while working on https://github.com/NixOS/nixpkgs/pull/276662, I tried it on my own PR. I ran:

```console
$ nix-shell -p nixpkgs-review --run "nixpkgs-review rev HEAD"
$ git -c fetch.prune=false fetch --no-tags --force https://github.com/NixOS/nixpkgs master:refs/nixpkgs-review/0
remote: Enumerating objects: 200, done.
remote: Counting objects: 100% (139/139), done.
remote: Compressing objects: 100% (57/57), done.
remote: Total 92 (delta 73), reused 48 (delta 32), pack-reused 0 (from 0)
Unpacking objects: 100% (92/92), 13.61 KiB | 240.00 KiB/s, done.
From https://github.com/NixOS/nixpkgs
   31a69c3d855b..61ddb09cfaa7  master     -> refs/nixpkgs-review/0
$ git worktree add /home/vladimir/.cache/nixpkgs-review/rev-0752b57db97bb02e23eb0880855baa841ccee4b0-4/nixpkgs 61ddb09cfaa7424d7fc8e3040ccd5c8c6f875b15
Preparing worktree (detached HEAD 61ddb09cfaa7)
Updating files: 100% (43112/43112), done.
HEAD is now at 61ddb09cfaa7 Revert "syncthing-relay: 1.27.9 -> 1.27.10" (#340552)
$ nix-env --extra-experimental-features no-url-literals --option system x86_64-linux -f <nixpkgs> --nix-path nixpkgs=/home/vladimir/.cache/nixpkgs-review/rev-0752b57db97bb02e23eb0880855baa841ccee4b0-4/nixpkgs nixpkgs-overlays=/tmp/tmpjeha_iv6 -qaP --xml --out-path --show-trace --no-allow-import-from-derivation
$ git merge --no-commit --no-ff 0752b57db97bb02e23eb0880855baa841ccee4b0
Automatic merge went well; stopped before committing as requested
[...]

PR #416, opened 2024-09-09
fix: distinguish build failures from store failures

CyberShadow:
"nix-store --verify-path" exits with status 1 in various situations, including when the path does not exist (due to having failed to build) or when the checksum doesn't match the Nix database.

Since nixpkgs-review tries to report build failures specifically, use "nix store verify" instead with its "--no-contents" flag. This avoids classifying checksum mismatches as build failures, which can be misleading to nixpkgs-review users.

Though still an abnormal situation that might be useful to report, we now ignore store path errors; these are generally an abnormal situation, and detecting them is out-of-scope for nixpkgs-review.  Should it become clearer that it would be useful to report them, it could be done by removing --no-contents and checking "nix store verify"'s exit status, which indicates the nature of the problem.

Fixes https://github.com/Mic92/nixpkgs-review/issues/415

### DIFF
--- a/nixpkgs_review/nix.py
+++ b/nixpkgs_review/nix.py
@@ -33,7 +33,17 @@ class Attr:
             return self._path_verified
 
         res = subprocess.run(
-            ["nix-store", "--verify-path", self.path], stderr=subprocess.DEVNULL
+            [
+                "nix",
+                "--extra-experimental-features",
+                "nix-command",
+                "store",
+                "verify",
+                "--no-contents",
+                "--no-trust",
+                self.path,
+            ],
+            stderr=subprocess.DEVNULL,
         )
         self._path_verified = res.returncode == 0
         return self._path_verified


CyberShadow:
I tested the two approaches with this script:

```bash
#!/usr/bin/env bash
set -eEuo pipefail

expr='
let
  name = builtins.getEnv "name";  # https://github.com/NixOS/nix/issues/2678
  pkgs = import <nixpkgs> {};
  pkg = pkgs.stdenv.mkDerivation {
    inherit name;
    dontUnpack = true;
    buildCommand = if name == "broken" then "false" else "mkdir $out";
  };
in {
  path = "${pkgs.lib.getOutput "out" pkg}";
  inherit pkg;
}
'

for f in good corrupted broken ; do
  name="$f" \
  nix-build \
    -A pkg \
    --expr "$expr" || true

  path=$(
    name="$f" \
    nix \
      --extra-experimental-features nix-command \
      eval \
      --expr "$expr" \
      path \
      --impure \
      --raw
  )
  if [[ "$f" == corrupted ]] ; then
    sudo touch     "$path"/oops.txt
    sudo chmod 666 "$path"/oops.txt
  fi

  status=0
  # old method:
  # nix-store --verify-path "$path" || status=$?
  # new method:
  nix --extra-experimental-features nix-command store verify --no-contents --no-trust "$path" || status=$?
  echo "$f: $status"
done
```

The "old method" fails on `broken` and `corrupted`. The "new method" fails only on "broken", as desired.

Mic92:
@mergify queue

Mic92:
Thanks!

PR #419, opened 2024-09-22
Feat: add support for multi-systems

GaetanLepage:
This PR addresses #400 by allowing to evaluate a PR/revision for multiple systems at once.
You can now provide the `--systems` option with a list of systems to evaluate:
- `--systems current` (default): legacy behavior
- `--systems "x86_64-linux aarch64-darwin"`: evaluate for both intel linux and ARM darwin
- `--systems all`: evaluate on all four main platforms

### Progress

- [x] Overall refactoring, addition of the necessary abstractions
- [x] Adapted all "output" features (`report.json`, `report.md`, stdout...)
- [x] Support for `--eval ofborg`
- [x] Support for `--eval local` (sequential)
- [ ] ~~Support for `--eval local` (optionally parallel)~~ -> Let's do this in a following PR. This one is already big enough.
- [x] Flesh out documentation
- [x] Final code polishing

> [!Note]
>
> This is still a WIP. ~~The tool works fine when using ofborg eval but I have not yet adapted to local evaluation.~~
> The code is quite dirty for now and will require more polish ~~once everything will be working~~.
> Also, I am fully open to suggestions regarding the solution and coding style.

I am eager to receive some feedback !

### Technical details

- Basically, adding this feature requires a pretty heavy refactoring.
Every place where there was a list of attributes/packages/reports, we now need a `dict` mapping each `system` to this list.
- I also needed to change the code for the nix shell to build as it now combines the packages to build for all platforms in the final attribute list.
- The current `--system` option has been softly deprecated (warning + forwarding the value to `--systems`) to avoid breaking the workflow of existing users.
- **Note:** I also had to change the layout of the `report.json`.

cc @Mic92 @natsukium

PR #422, opened 2024-10-02
Feat: add basic parallel support for listing packages

GaetanLepage:
When ofborg's evaluation results are not available or when the user specifies `--eval local`, local evaluation is used to list packages affected by a PR.
This process is quite time-consuming and could be done in parallel when evaluating for multiple systems.

Ultimately, the performance of [nix-eval-jobs](https://github.com/nix-community/nix-eval-jobs) could be leveraged to perform this task.
Unfortunately, as it is currently, `nix-eval-jobs` cannot be used as a drop-in replacement in this code base.

As an intermediary situation, I suggest to ship a simple parallel solution.
This approach can be very memory-intensive.
Running it with 4 parallel processes (one for each system) consumes approximately 55GB of RAM.
Nonetheless, it unsurprisingly reduces the time spent on evaluation almost linearly.

I did a naive test running `nixpkgs-review --systems all` (4 systems) on a PR I had already reviewed (all build artifacts were cached on my system).
The total wall-time was *5:17* with 4 parallel processes and *16:51* with a single process.

An important decision to make is about how to parametrize the number of threads.
#419 introduced `--num-procs-eval` which is currently used to control how many `nix eval` commands were ran in parallel.
Although it could be convenient to use this single flag for both processes, the implication in terms of memory usage are significantly different.

How would you see the configuration for this new parallel feature ?
- A new flag ? If so, do we keep the current name ?
- Having two flags ? (Renaming the old one in the meantime ?)

### DIFF
--- a/nixpkgs_review/review.py
+++ b/nixpkgs_review/review.py
@@ -108,7 +108,7 @@ class Review:
         skip_packages_regex: list[Pattern[str]] = [],
         checkout: CheckoutOption = CheckoutOption.MERGE,
         sandbox: bool = False,
-        n_procs_eval: int = 1,
+        num_parallel_evals: int = 1,
     ) -> None:
         self.builddir = builddir
         self.build_args = build_args
@@ -141,7 +141,7 @@ class Review:
         self.build_graph = build_graph
         self.nixpkgs_config = nixpkgs_config
         self.extra_nixpkgs_config = extra_nixpkgs_config
-        self.n_procs_eval = n_procs_eval
+        self.num_parallel_evals = num_parallel_evals
 
     def worktree_dir(self) -> str:
         return str(self.builddir.worktree_dir)
@@ -186,7 +186,7 @@ class Review:
             self.builddir.nix_path,
             self.systems,
             self.allow,
-            n_procs=self.n_procs_eval,
+            n_threads=self.num_parallel_evals,
         )
 
         if reviewed_commit is None:
@@ -199,7 +199,7 @@ class Review:
             self.builddir.nix_path,
             self.systems,
             self.allow,
-            n_procs=self.n_procs_eval,
+            n_threads=self.num_parallel_evals,
             check_meta=True,
         )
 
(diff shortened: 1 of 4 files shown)


Theghost256 (approved):

Mic92:
If you eval the whole of nixpkgs for different systems than this approach is probably even faster because it overlaps less duplicate computation compared to nix-eval-jobs.
Each system is independent from the others.

GaetanLepage:
> If you eval the whole of nixpkgs for different systems than this approach is probably even faster because it overlaps less duplicate computation compared to nix-eval-jobs.
> Each system is independent from the others.

Yes, I have been using it lately and it works fine.

However, I am still wondering what we should do w.r.t the user interface.
Should we have two flags to separate the `nix-env` parallelization and the `nix eval` one ?

Mic92:
@mergify rebase

nixpkgs_review/review.py
```diff
@@ -449,7 +441,30 @@ def list_packages(
...
+) -> dict[System, list[Package]]:
+    list_packages_partial = partial(
+        _list_packages_system,
+        nix_path=nix_path,
+        allow=allow,
+        check_meta=check_meta,
+    )
+
+    args: list[System] = list(systems)
+    with mp.Pool(n_procs) as pool:
```
Mic92, line 462: Can we change this to https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.ThreadPoolExecutor
processes take longer to spawn than threads.

Mic92: Probably should also be changed in the other place.

GaetanLepage: Done !

Mic92:
@mergify queue

GaetanLepage:
Is it stuck ?

Mic92:
Yeah. Mergify is buggy lately.

Mic92:
@mergify queue

GaetanLepage:
> Yeah. Mergify is buggy lately.

Ok, thanks for the careful review :)

Issue #267, opened 2022-04-18
respect shallow/grafted nixpkgs repositories


xfnw:
if you attempt to nixpkgs-review pr on a shallow copy of nixpkgs, it will try to download the entire ~4gb of git history

Mic92:
I don't see a way around this. It currently uses `git fetch` to download the history of both master and the PR and than needs to perform a merge of one into the other

SuperSandro2000:
> it will try to download the entire ~4gb of git history

For me the entire history is just 1.3 GB big.

We could probably modify the fetch commands to only fetch what we really need but it is a pain to work with shallow repositories since certain things done with the history will just fail in some situations and executing multiple fetches for every user is probably slowing down things more.

xfnw:
> For me the entire history is just 1.3 GB big.

huh... do you have some sort of compression enabled? mine seems to be a lot larger
```
[xfnw@feesh:~]$ du -sh nixpkgs
3.7G	nixpkgs

[xfnw@feesh:~]$ du -sh nixpkgs/.git
3.6G	nixpkgs/.git
```

SuperSandro2000:
Try

```
du -sh .git
git gc --aggressive
git gc
git repack -Ad
git prune
du -sh .git
```

lolbinarycat:
@SuperSandro2000 
> We could probably modify the fetch commands to only fetch what we really need but it is a pain to work with shallow repositories since certain things done with the history will just fail in some situations and executing multiple fetches for every user is probably slowing down things more.

what about an arbitrarily large but still much smaller than the whole repository amount?

for example, fetching only the last 10000 commits?

waiting 10 minutes is still a lot better than waiting 2 hours, and people using a shallow clone are likely to have a a slow network connection anyways.

PR #426, opened 2024-10-27
Support nixpkgs shallow clone and make use of github pull/<pr>/merge ref

azuwis:
Fix https://github.com/Mic92/nixpkgs-review/issues/267

### DIFF
--- a/tests/test_pr.py
+++ b/tests/test_pr.py
@@ -26,8 +26,8 @@ def test_pr_local_eval(helpers: Helpers, capfd: pytest.CaptureFixture) -> None:
             f.write("foo")
         subprocess.run(["git", "add", "."])
         subprocess.run(["git", "commit", "-m", "example-change"])
-        subprocess.run(["git", "checkout", "-b", "pull/1/head"])
-        subprocess.run(["git", "push", str(nixpkgs.remote), "pull/1/head"])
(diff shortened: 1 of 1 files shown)


nixpkgs_review/review.py
```diff
@@ -272,15 +274,19 @@ def build_pr(self, pr_number: int) -> dict[System, list[Attr]]:
...
-            pr["base"]["ref"],
-            f"pull/{pr['number']}/head",
-        )
 
         if self.checkout == CheckoutOption.MERGE:
-            base_rev = merge_rev
+            base_rev, pr_rev = fetch_refs(
+                self.remote,
+                pr["base"]["ref"],
+                f"pull/{pr['number']}/merge",
```
Mic92, line 282: In fact there is a race condition with this ref sometimes as github needs some time to compute it. But this is probably only relevant for CIs. So we should be unlikely to when humans run nixpkgs-review (except for @SuperSandro2000 maybe)

Mic92:
@mergify queue

azuwis:
I've found one regression, if the PR is merged, `pull/<PR>/merge` will be gone, and won't be able to run nixpkgs-review against a merged PR after this.

Mic92:
Can we detect if `/pull/<PR>/merge` is available and otherwise fallback?

PR #427, opened 2024-11-05
nix/review_shell.nix: use ignoreFileOutputs whenever available

### LINKED ISSUE #408: Review shell fails when there are single-file inputs.
Some things that get covered by `nixpkgs-review` install with something like `touch $out`, which can result in this:

```
       > error: The store path /nix/store/b5q8bljrad33lmmypn4dd2firrjg1rs5-equal-contents-builddotnetmodule-sets-assemblycopyrightattribute-with-structured-attributes is a file and can't be merged into an environment using pkgs.buildEnv! at /nix/store/77my1q03gyb36mhs0sij87yr710w9gha-builder.pl line 122.
```

I've fixed a couple by changing them to `mkdir $out`, but occasionally there's a reason to have content in the output file, so perhaps this should be supported?

ShamrockLee:
This PR depends on NixOS/nixpkgs#353752 and fixes #408 partially.


ShamrockLee:
Considering that the single-file-output error is a potential way to enforce the correctness of `meta.outputsToInstall`, we may want to preserve such behavior while proving options to ignore those single-file outputs.

I prepared the implementation to support `nixpkgs-review --ignore-single-file-outputs` locally. I could push it up if we agree that this should be supported through a flag.

@Mic92, what do you think?

nixpkgs_review/nix/review-shell.nix
```diff
@@ -26,11 +26,15 @@ let
     in
     map (attrString: lib.attrByPath (lib.splitString "." attrString) null system-pkg) system-attrs;
   attrs = lib.flatten (lib.mapAttrsToList extractPackagesForSystem (import attrs-path));
-  env = local-pkgs.buildEnv {
-    name = "env";
-    paths = attrs;
-    ignoreCollisions = true;
-  };
+  env =
+    local-pkgs.buildEnv {
```
corngood, line 30: I think this needs parens around the attrset update, so `ignoreSingleFileOutputs` makes it into `buildEnv`.

It does seem to work okay once that's changed.

corngood (changes requested):
I'm not sure if it's something to do with the change in evaluation methods, but lately I'm getting a lot more file conflicts due to "test.*" being included in attrs.nix, so it would be nice to get this in.

ShamrockLee:
@corngood Thank you for rewiewing!

Do you think it's preferred to have `ignoreSingleFileOutputs = true` by default (the current implementation), or provide a flag `--ignore-single-file-outputs` to opt in explicitly?

corngood:
> Do you think it's preferred to have ignoreSingleFileOutputs = true by default (the current implementation), or provide a flag --ignore-single-file-outputs to opt in explicitly?

My gut feeling is that it should be a default, because people often run `nixpkgs-review` without knowledge of what's actually going to be built, and there are a bunch of things in nixpkgs that will break if they do happen to be built as part of a review.  For example, I hit this on every update to `dotnet-sdk`, because of things in `tests.dotnet`.

However, you did mention this:

> Considering that the single-file-output error is a potential way to enforce the correctness of meta.outputsToInstall

and I don't really understand what you mean by it.  Maybe you can give an example?

ShamrockLee:
> but lately I'm getting a lot more file conflicts due to "test.*" being included in attrs.nix

Is it the default behavior now? I thought it would be opt-in.

corngood:
> Is it the default behavior now? I thought it would be opt-in.

It's the top level `tests` attribute, not e.g. `passthru.tests`.  It seems to be something to do with the github eval.  See this review run:

https://github.com/NixOS/nixpkgs/pull/389338#issuecomment-2719253926

- tests.dotnet.final-attrs.check-output
- tests.dotnet.final-attrs.output-matches-const
- tests.dotnet.final-attrs.override-has-no-effect

etc

That one was done with github eval.  Below that I posted a darwin review, which used local eval, and it _didn't_ build anything in `tests`.

corngood:
> That one was done with github eval. Below that I posted a darwin review, which used local eval, and it didn't build anything in tests.

I looked into this a bit, and I think it's just the difference between local eval using `nix-env -qaP` and the eval action using `release-attrpaths-superset.nix`.

I didn't realise they were so drastically different (essentially opt-in vs opt-out).

ShamrockLee:
Thank you for pinging. This PR slipped under my radar.

Considering the default behavior of nixpkgs-review to evaluate `pkgs.tests`, let's set `ignoreSingleFileOutputs = true` by default.

I'm adding `--check-single-file-outputs` to allow `ignoreSingleFileOutputs`.

ShamrockLee:
> I needed to get it to find more than 50 attributes, due to `packages = if builtins.length attrs > 50 then [ env ] else attrs;`.
> 
> With <= 50 attributes it works even with single files when they are passed directly into `mkShell`. Because of that I think maybe it's not worth adding `--check-single-file-outputs`.

Thank you for finding out this isssue. I fixed the condition by using `[ env ]` whenever `ignoreSingleFileOutputs` is `false`.

ShamrockLee:
Just updated to call `nix_build()` and `nix_shell()` with keyword arguments instead.

philiptaron:
I would like to see this PR merged because it would help me with such PRs as:

- https://github.com/NixOS/nixpkgs/pull/437584

How can I help get this to mergeable status?

corngood:
Personally I think we should just set `ignoreSingleFileOutputs` unconditionally and remove `--check-single-file-outputs`.  The flag was added out of an abundance of caution to maintain the old behaviour:

> Considering that the single-file-output error is a potential way to enforce the correctness of meta.outputsToInstall, we may want to preserve such behavior while proving options to ignore those single-file outputs.

However the old behaviour was dependent on the number of packages being built relative to that undocumented magic threshold (50), so nobody could really depend on it anyway.  At least that's how I understand it.

@ShamrockLee what do you think about that?  If you'd prefer I can make a separate PR with what I'm suggesting, but it's really just going back to how this PR started.

ShamrockLee:
I changed my mind. It's so hard to get the command-line flag to pass the tests.

I'll reset the feature branch to specify `ignoreSingleFileOutputs` whenever available.

corngood (approved):
LGTM.  Thanks!

PR #429, opened 2024-11-17
chunked local evaluation

GaetanLepage:
See https://github.com/NixOS/nixpkgs/pull/356023

- **default.nix: python3.pkgs -> python3Packages**
- **chunked eval**

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -11,20 +11,20 @@ let
     withNom
     && (builtins.tryEval (builtins.elem buildPlatform.system pkgs.ghc.meta.platforms)).value or false;
 in
-python3.pkgs.buildPythonApplication {
+python3Packages.buildPythonApplication {
   name = "nixpkgs-review";
   src = ./.;
   format = "pyproject";
-  nativeBuildInputs = [ installShellFiles ] ++ lib.optional withAutocomplete python3.pkgs.argcomplete;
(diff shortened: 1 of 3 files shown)


infinisil:
Also check out https://github.com/tweag/nixpkgs/tree/gha-eval/ci/eval#nixpkgs-ci-evaluation, that's the latest state of the above-linked PR

Mic92:
Eval got merged. We can also start using the json.

JohnRTitor:
I would add that instead of fetching @ofborg's evaluation result, if available it should fetch our new CI evaluator. https://github.com/NixOS/nixpkgs/actions/runs/11947073470/job/33302502332?pr=357763

Mic92:
> I would add that instead of fetching @ofborg's evaluation result, if available it should fetch our new CI evaluator. https://github.com/NixOS/nixpkgs/actions/runs/11947073470/job/33302502332?pr=357763

So much todo and so little time :). If you want to take a step at it, I wouldn't mind. We can even check for both ofborg and the github one so it also works for older PRs that haven't been updated. However we don't have that json for master yet.

Mic92:
Superseded by #693: the package listing runs through nix-eval-jobs, whose per-worker memory limit gives the same effect as chunking.

PR #432, opened 2024-11-23
Add --git-config flag

ShamrockLee:
This PR adds a `--git-config` flag to pass additional Git configurations to the Git executable via the `-c name value` arguments. This flag can be specified multiple times, preserving spaces inside the configuration values in each specification.

This flag allows running `nixpkgs-review` with trusted users (usually `root`) but with the original user's Git configurations (such as `user.name` and `user.email`), which helps gain builder access for multi-system reviews.

### DIFF
--- a/nixpkgs_review/cli/__init__.py
+++ b/nixpkgs_review/cli/__init__.py
@@ -164,6 +164,12 @@ def common_flags() -> list[CommonFlag]:
             "--build-args", default="", help="arguments passed to nix when building"
         ),
         CommonFlag(
+            "--git-config",
+            action="append",
+            default=[],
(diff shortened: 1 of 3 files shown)


Mic92:
We could just do this:

```
os.environ["GIT_AUTHOR_NAME"] = "nixpkgs-review"
os.environ["GIT_AUTHOR_EMAIL"] = "nixpkgs-review@example.com"
os.environ["GIT_COMMITTER_NAME"] = "nixpkgs-review"
os.environ["GIT_COMMITTER_EMAIL"] = "nixpkgs-review@example.com"
```

Mic92:
https://github.com/Mic92/nixpkgs-review/pull/528

PR #434, opened 2024-12-12
flake update

GaetanLepage:
- **flake.lock: Update**
- **Reformat code base**

### DIFF
--- a/nixpkgs_review/cli/comments.py
+++ b/nixpkgs_review/cli/comments.py
@@ -1,4 +1,5 @@
 import argparse
+import string
 from dataclasses import dataclass, field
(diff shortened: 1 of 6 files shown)


nixpkgs_review/cli/comments.py
```diff
@@ -8,45 +8,45 @@
 
 
 def comments_query(pr: int) -> str:
-    return """
-{
-    repository(owner: "NixOS", name: "nixpkgs") {
-        pullRequest(number: %d) {
```
Mic92, line 14: Maybe we want to keep the %d here actually and make ruff ignore this block.

PR #435, opened 2024-12-12
Fetch eval result comparison from GH workflow artifact

### LINKED ISSUE #433: GHA eval support
Since https://github.com/NixOS/nixpkgs/pull/359704, there is an evaluation output available as an artifact in the `Eval / Process (pull_request_target)` check. It would be nice to reuse it to skip local evaluation.

GaetanLepage:
`nixpkgs-review` can now retrieve the list of rebuilds to do in the GitHub _Eval_ action artifact.
Leverages https://github.com/NixOS/nixpkgs/pull/363751 to resolve #433.

- **rename OFBORG_PLATFORMS to PLATFORMS**
- **Fetch eval result comparison from GH workflow artifact**


GaetanLepage:
Should we add a bit more verbosity ?
For example, telling which eval strategy is currently being tried ?

ShamrockLee:
> Should we add a bit more verbosity?
> For example, telling which eval strategy is currently being tried.

It's a good idea. One thing I like about nixpkgs-review is that it always prints out what it's doing in a manually reproducible way. This way, it automates the review task without becoming another black box that reviewers need to debug.

GaetanLepage:
Added basic logging:
image

Feel free to make additional suggestions.

GaetanLepage:
Hmm, it's not working currently.

Mic92:
@mergify refresh

Mic92:
@mergify queue

Mic92:
@mergify refresh

PR #440, opened 2024-12-22
local evaluation: use nixpkgs ci.eval parallel implementation

GaetanLepage:
This PR aims at improving local evaluation.

Currently, it uses a (single-threaded) `nix-env` call per system to evaluate.
This new patch leverages the recent [`ci.eval` parallel evaluation implementation in `nixpkgs` itself](https://github.com/NixOS/nixpkgs/tree/master/ci/eval).

Current limitations:
- Not respecting `--allow` as we used to (maybe it is possible...)
- It requires carefully tuning the `--max-jobs`, `--cores` and `chunkSize` parameters of `ci.full`
    We need to expose new options for those and provide sensible (automatic ?) default values.
- The listing of "Impacted packages" on `stdout` is now limited to listing all "rebuilds" (i.e. added + changed packages). Before, it was more detailed (_removed_, _changed_ and _added_ I think). This information is not available in the output of the `eval.compare` derivation. If needed, we could change the upstream implem...

---

## TODO

- [ ] Handle `--allow` user preferences
- [ ] Fix tests which all got broken by this patch
- [ ] (Automatically) set **sensible** default values for the `eval.full` parameters (`--max-jobs`, `--cores` and `--arg chunkSize`)

cc @Mic92 @zowoq @khaneliman

### DIFF
--- a/nixpkgs_review/eval_ci.py
+++ b/nixpkgs_review/eval_ci.py
@@ -1,0 +1,75 @@
+from pathlib import Path
+
+from .utils import System, sh
+
+
+def _ci_command(
+    worktree_dir: Path,
+    command: str,
+    output_dir: str,
+    options: dict[str, str] | None = None,
+    args: dict[str, str] | None = None,
+) -> None:
+    cmd: list[str] = [
+        "nix-build",
+        str(worktree_dir.joinpath(Path("ci"))),
+        "-A",
+        f"eval.{command}",
+    ]
+    if options is not None:
+        for option, value in options.items():
+            cmd.extend([option, value])
+
+    if args is not None:
+        for arg, value in args.items():
+            cmd.extend(["--arg", arg, value])
+
+    cmd.extend(["--out-link", output_dir])
+    sh(cmd, capture_output=True)
+
+
+def local_eval(
+    worktree_dir: Path,
+    systems: set[System],
+    max_jobs: int,
+    n_cores: int,
+    chunk_size: int,
+    output_dir: str,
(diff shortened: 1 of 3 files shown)


infinisil:
You should be able to get the same speed as the legacy `nix-env` by setting the chunk size to like 1000000 (there's about 100000 attributes right now) at least, but to get the best speed I don't have a better answer than just to try a bunch of different numbers for the chunk size

Mic92:
It feels like we need to solve this in nix. i.e. track RAM usage there and restart the interpreter after a threshold is breached.

Mic92:
> So, what should we do about this ? While I find this new implem more satisfying, cleaner and allowing finer control over the evaluation process, one must admit that it is significantly less performant.

Maybe the new method could be a different branch for the time being. Some people might have machines with enough resources, where it actually speeds up things. I am wondering if we can leverage `libnix` somehow to restart evaluation when it consumes to much RAM.

infinisil:
> While I find this new implem more satisfying, cleaner and allowing finer control over the evaluation process, one must admit that it is significantly less performant.

Did you try my suggestion of having a single huge chunk, so that it should effectively be the same single-threaded performance as before?

infinisil:
I can only imagine that nixpkgs-review doesn't do the same eval then

Mic92:
Local evaluation now uses nix-eval-jobs with parallel, memory-bounded workers (#693), which covers this without depending on nixpkgs' ci/eval. Thanks!

PR #451, opened 2025-01-05
Wait for GHA eval results instead of falling back to local evaluation

GaetanLepage:
The default evaluation mode (`github`) attempts at fetching the evaluation results from the GitHub actions result.
Those will only be available a few minutes after the PR has been opened as they are not trivial to compute.

Currently, if the results are not available right away, `nixpkgs-review` falls back to local evaluation.
This PR changes this behavior. Now it will keep re-trying until the results are available.

Note: It is still possible to force local evaluation by using `--eval local`.

**Motivation:** Local evaluation is very resource-intensive (especially RAM). It should be an opt-in behavior.
This new approach is more predictable.

Questions:
- [x] Does this work (at all) ?
- [x] Is 10s a good value for the timeout ? -> Seems good, yes.
- [x] Should we add a maximum amount of retries ?
  -> I have set a maximum waiting time of 10min:

image

- [x] ~~Handle `Ctrl-C` correctly (i.e. properly delete the worktree on exit)~~
  -> There is no worktree at this point. Just quit gracefully.


nixpkgs_review/review.py
```diff
@@ -296,14 +297,20 @@ def build_pr(self, pr_number: int) -> dict[System, list[Attr]]:
...
+        if self.use_github_eval:
+            assert all(system in PLATFORMS for system in self.systems)
+            print("-> Fetching eval results from GitHub actions")
 
-            if packages_per_system is not None:
-                print("-> Successfully fetched rebuilds: no local evaluation needed")
+            packages_per_system = self.github_client.get_github_action_eval_result(pr)
+            timeout: int = 30
+            while packages_per_system is None:
+                print(f"...Results are not (yet) available. Retrying in {timeout}s")
```
Mic92, line 307: Maybe just print this message just once and just add a dot add the end each time it's not finished.

nixpkgs_review/review.py
```diff
@@ -296,14 +297,20 @@ def build_pr(self, pr_number: int) -> dict[System, list[Attr]]:
...
-            print("-> Attempting to fetch eval results from GitHub actions")
-            packages_per_system = self.github_client.get_github_action_eval_result(pr)
+        if self.use_github_eval:
+            assert all(system in PLATFORMS for system in self.systems)
+            print("-> Fetching eval results from GitHub actions")
 
-            if packages_per_system is not None:
-                print("-> Successfully fetched rebuilds: no local evaluation needed")
+            packages_per_system = self.github_client.get_github_action_eval_result(pr)
+            timeout: int = 30
```
Mic92, line 305: I'd say 10s here. We have a fair amount of requests we can do our token.

nixpkgs_review/review.py
```diff
@@ -296,14 +297,22 @@ def build_pr(self, pr_number: int) -> dict[System, list[Attr]]:
...
-            print("-> Attempting to fetch eval results from GitHub actions")
-            packages_per_system = self.github_client.get_github_action_eval_result(pr)
-
-            if packages_per_system is not None:
-                print("-> Successfully fetched rebuilds: no local evaluation needed")
+        if self.use_github_eval:
+            assert all(system in PLATFORMS for system in self.systems)
+            print("-> Fetching eval results from GitHub actions")
 
+            packages_per_system = self.github_client.get_github_action_eval_result(pr)
```
Mic92, line 304: I believe this also returns None if the evaluation failed. However in this case we shouldn't wait as we would wait forever.

nixpkgs_review/review.py
```diff
@@ -296,14 +297,22 @@ def build_pr(self, pr_number: int) -> dict[System, list[Attr]]:
...
+            if packages_per_system is None:
+                timeout: int = 10
+                print(f"...Results are not (yet) available. Retrying in {timeout}s")
+                while packages_per_system is None:
+                    print(".", end="")
```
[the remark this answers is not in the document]
Mic92: I think it's line buffered. Try flushing sys.stdout.flush()

nixpkgs_review/review.py
```diff
@@ -296,14 +297,23 @@ def build_pr(self, pr_number: int) -> dict[System, list[Attr]]:
...
+            assert all(system in PLATFORMS for system in self.systems)
+            print("-> Fetching eval results from GitHub actions")
 
+            packages_per_system = self.github_client.get_github_action_eval_result(pr)
+            if packages_per_system is None:
+                timeout: int = 10
+                print(f"...Results are not (yet) available. Retrying in {timeout}s")
+                while packages_per_system is None:
+                    sys.stdout.flush()
+                    print(".")
```
Mic92, line 310: Flush after print.

PR #460, opened 2025-02-02
Show `--skip-package` in the report

PerchunPak:
CC @GaetanLepage


PerchunPak:
Unnable to test, I broke my nixpkgs clone so I am clonning it all over again

PerchunPak:
> What would have been great is to list packages that have actually been skipped.
> This implies keeping track of the packages filtered in the evaluation process.

That requires a refactor, and I don't really want to do it

PR #472, opened 2025-02-24
Fix Review.build_commit merge behavior

B4dM4n:
The `self.checkout` test is the wrong way round, so `nix-review rev` always does a `git checkout` and not merge the given commit.


Mic92:
@mergify queue

PR #473, opened 2025-02-28
Switch to `nix-eval-jobs` for local evaluation

B4dM4n:
This PR replaces all calls to `nix-env -qaP` and `nix eval` with calls to [nix-eval-jobs](https://github.com/nix-community/nix-eval-jobs). This allows the user to control the memory consumption of `nixpkgs-review` very accurately.

Each commit in this PR can be tested and reviewed individually.

With `nix-eval-jobs`, the user now has to options to control the memory usage of `nixpkgs-review`:
 - `--num-parallel-evals`: the same flag that existed before and controlled the number of parallel `nix-env` calls, is now used for the `--worker` argument to `nix-eval-jobs` (default 1)
 - `--max-memory-size`: this is passed directly to `nix-eval-jobs` and controls the amount of memory in MB, that each worker can consume (default 4096).

Due to the internals of `nix-eval-jobs`, a worker can shortly consume more memory than specified by `--max-memory-size`, if the evaluation of the current attribute is large (e.g. nixos tests). But once the evaluation of that attribute is finished, the worker restarts and the memory is freed.

The `default.nix` currently includes an unreleased `nix-eval-jobs` version, because the `--apply` flag is not released yet.

To reduce the memory usage further, the invoked `nix build` command will now receive the derivations to build directly, instead of evaluating all attributes itself again (which can also consume multiple GB of RAM).

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -10,6 +10,18 @@ let
   withNom' =
     withNom
     && (builtins.tryEval (builtins.elem buildPlatform.system pkgs.ghc.meta.platforms)).value or false;
+
+  # at least v2.26.0 (currently unreleased) is required for the '--apply' flag
+  nix-eval-jobs =
+    if lib.versionOlder pkgs.nix-eval-jobs.version "2.26.0" then
+      callPackage (fetchFromGitHub {
+        owner = "nix-community";
+        repo = "nix-eval-jobs";
+        rev = "6d4fd5a93d7bc953ffa4dcd6d53ad7056a71eff7";
+        hash = "sha256-1dZLPw+nlFQzzswfyTxW+8VF1AJ4ZvoYvLTjlHiz1SA=";
+      }) { nix = nixVersions.nix_2_25; }
+    else
+      pkgs.nix-eval-jobs;
 in
 python3Packages.buildPythonApplication {
   name = "nixpkgs-review";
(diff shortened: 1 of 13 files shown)


default.nix
```diff
@@ -46,6 +58,7 @@ python3Packages.buildPythonApplication {
         [
           pkgs.nixVersions.stable or nix_2_4
```
SuperSandro2000, line 59:
```suggestion
          pkgs.nixVersions.stable
```
I don't think that is necessary anymore

flake.nix
```diff
@@ -3,6 +3,7 @@
 
   inputs = {
     nixpkgs.url = "github:NixOS/nixpkgs/nixpkgs-unstable";
+
```
SuperSandro2000, line 6:
```suggestion
```

nixpkgs_review/nix.py
```diff
@@ -318,33 +346,29 @@ def nix_build(
...
 
     command = [
         build_graph,
         "build",
-        "--file",
-        REVIEW_SHELL,
-        "--nix-path",
-        nix_path,
         "--extra-experimental-features",
-        "nix-command" if allow.url_literals else "nix-command no-url-literals",
```
SuperSandro2000, line 342: Do we need to remove allow.url_literals now completely?

Mic92:
The nix-eval-jobs parts landed via #595, #568 and #690. Replacing nix-env -qaP for the local rebuild diff is still open but would need a fresh PR against current main.

PR #474, opened 2025-03-09
report: multithread write_error_logs, avoid hammering cache.nixos.org

pbsds:


nixpkgs_review/report.py
```diff
@@ -56,44 +57,106 @@ def ensure(self) -> Path:
...
+    *,
+    max_workers: int | None = 1,
+) -> None:
+    logs = LazyDirectory(directory.joinpath("logs"))
+    results = LazyDirectory(directory.joinpath("results"))
+    failed_results = LazyDirectory(directory.joinpath("failed_results"))
+
+    extra_nix_log_args = []
+
+    # filter https://cache.nixos.org from acting as build-log subtituter
```
Mic92, line 107:
```suggestion
    # filter https://cache.nixos.org from acting as build-log substituter
```

nixpkgs_review/report.py
```diff
@@ -56,44 +57,106 @@ def ensure(self) -> Path:
...
+    # to avoid hammering it
+    # IDEA: also add the remote builders if user has not already configured this
+    # TODO: should this option respect '--build-args'? 'nix log' accepts most, but not all
+    substituters = get_nix_config("substituters").get("substituters")
+    if substituters is not None:
+        extra_nix_log_args += [
+            "--option",
+            "substituters",
```
Mic92, line 115: I don't recall the details, but as an untrusted user your are able to make the list shorter than the default substituters from /etc/nix/nix.conf?

Mic92: Looks like this should work in principle?

```
            auto setSubstituters = [&](Setting<Strings> & res) {
                if (name != res.name && res.aliases.count(name) == 0)
                    return false;
                StringSet trusted = settings.trustedSubstituters;
                for (auto & s : settings.substituters.get())
                    trusted.insert(s);
                Strings subs;
                auto ss = tokenizeString<Strings>(value);
                for (auto & s : ss)
                    if (trusted.count(s))
                        subs.push_back(s);
                    else if (!hasSuffix(s, "/") && trusted.count(s + "/"))
                        subs.push_back(s + "/");
                    else
                        warn("ignoring untrusted substituter '%s', you are not a trusted user.\n"
                             "Run `man nix.conf` for more information on the `substituters` configuration option.", s);
                res = subs;
                return true;
            };
```

pbsds: Both the [documentation](https://nix.dev/manual/nix/2.26/command-ref/conf-file.html#conf-substituters) and the [full code](https://github.com/NixOS/nix/blob/a047dec120672d00e069bacf10ffdda420fd1048/src/libstore/daemon.cc#L221) (my first time vieweing the nix c++ code, not too shabby) indicate that the trusted-subtituters (TIL) only serves as a whitelist.

pbsds:
Any other inputs @Mic92? @GaetanLepage could you also have a look?

pbsds:
I've tested that `--option substituters ""` causes `nix log` to fail to retrieve the build log if it is not in the local store. `subprocess.run` does not go through any shell, so no additional escaping is needed. I've also been using this on my big-parallel builder when checking larger rebuilds

PR #482, opened 2025-04-07
Add example to --systems help

roberth:
Without it, the syntax is unclear. For instance, I tried commas, and that didn't work.

### DIFF
--- a/nixpkgs_review/cli/__init__.py
+++ b/nixpkgs_review/cli/__init__.py
@@ -219,7 +219,7 @@ def common_flags() -> list[CommonFlag]:
             "--systems",
             type=str,
             default="current",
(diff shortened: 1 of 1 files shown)


nixpkgs_review/cli/__init__.py
```diff
@@ -219,7 +219,7 @@ def common_flags() -> list[CommonFlag]:
             "--systems",
             type=str,
             default="current",
-            help="Nix 'systems' to evaluate and build packages for",
+            help="Nix 'systems' to evaluate and build packages for (e.g. 'x86_64-linux aarch64-darwin')",
```
pbsds, line 222:
```suggestion
            help="Nix 'systems' to evaluate and build packages for (e.g. 'all' or 'x86_64-linux aarch64-darwin')",
```

PR #487, opened 2025-05-04
Print out log directory at the end of the run

booxter:
This is helpful when running with --no-shell where there's no "paused"
context of nix-shell to assume. The link is clickable.

(Yes, the user sees git workspace removed at the end of the run but it's
not very convenient to copy the root of nixpkgs checkout, then append
`logs` manually.)

Signed-off-by: Ihar Hrachyshka <ihar.hrachyshka@gmail.com>

### DIFF
--- a/nixpkgs_review/report.py
+++ b/nixpkgs_review/report.py
@@ -1,5 +1,6 @@
 import json
 import os
+import socket
 import subprocess
 from collections.abc import Callable
 from concurrent.futures import ThreadPoolExecutor
@@ -8,7 +9,7 @@ from re import Pattern
 from typing import Literal
 
 from .nix import Attr
-from .utils import System, info, link, skipped, system_order_key, warn
+from .utils import System, info, link, skipped, system_order_key, to_link, warn
 
 
 def print_number(
@@ -208,6 +209,12 @@ def order_reports(reports: dict[System, SystemReport]) -> dict[System, SystemRep
     )
 
 
+# https://gist.github.com/egmontkob/eb114294efbcd5adb1944c9f3cb5feda
+def to_file_uri(path: Path) -> str:
+    """Convert a path to a file URI, including hostname."""
+    return f"file://{socket.gethostname()}{path.absolute()}"
+
+
 class Report:
     def __init__(
         self,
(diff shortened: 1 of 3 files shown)


nixpkgs_review/utils.py
```diff
@@ -26,7 +26,10 @@ def wrapper(text: str) -> None:
 warn = color_text(31, file=sys.stderr)
 info = color_text(32)
 skipped = color_text(33)
-link = color_text(34)
+
+
+def link(uri: str) -> None:
+    color_text(34)(f"\u001b]8;;{uri}\u001b\\{uri}\u001b]8;;\u001b\\\n")
```
Mic92, line 32: The other PR only makes it a link if it is a TTY. This one should do the same.

booxter:
Will update to do it on tty only; and also may need to explore using hostname aware links for remote execution.

nixpkgs_review/utils.py
booxter: Sent this separately to fix the existing link: https://github.com/Mic92/nixpkgs-review/pull/489

nixpkgs_review/report.py
```diff
@@ -208,6 +209,12 @@ def order_reports(reports: dict[System, SystemReport]) -> dict[System, SystemRep
     )
 
 
+# https://gist.github.com/egmontkob/eb114294efbcd5adb1944c9f3cb5feda
+def to_file_uri(path: Path) -> str:
+    """Convert a path to a file URI, including hostname."""
+    return f"file://{socket.gethostname()}{path.absolute()}"
```
booxter, line 215: Including hostname give the desired behavior, e.g. in kitty:

```
Editing /home/booxter/.cache/nixpkgs-review/pr-405067-2/logs from build05
The command: ssh -o 'ControlPath=~/.ssh/kitty-rf-68330-%C' -o TCPKeepAlive=yes -o ControlPersist=yes -o BatchMode=yes linux-builder cat /home/booxter/.cache/nixpkgs-review/pr-405067-2/logs failed
cat: /home/booxter/.cache/nixpkgs-review/pr-405067-2/logs: Is a directory

Failed to download /home/booxter/.cache/nixpkgs-review/pr-405067-2/logs
```

Yes, it failed to edit it after all, but the main point is that it correctly detected that the URI is remote and tried to download it for edit. I believe I just need some adjustment to kitty to make this work too.

(In the next PR that will add links to each built package this won't be a problem anyway since these will point to actual files.)

booxter: @Mic92 The links are seen only on tty and they won't blow up the (visible) size of reports because links are hidden, they are activated by "clicking" them. And they will not go to Github.

PR #489, opened 2025-05-06
fix: Don't use link escape chars when no tty

booxter:


nixpkgs_review/report.py
```diff
@@ -328,7 +328,7 @@ def print_console(self, pr: int | None) -> None:
         if pr is not None:
             pr_url = f"https://github.com/NixOS/nixpkgs/pull/{pr}"
             info("\nLink to currently reviewing PR:")
```
SuperSandro2000, line 330:
```suggestion
            info("\nLink to currently reviewing PR:")
```

nixpkgs_review/report.py
```diff
@@ -328,7 +328,7 @@ def print_console(self, pr: int | None) -> None:
...
-            link(f"\u001b]8;;{pr_url}\u001b\\{pr_url}\u001b]8;;\u001b\\\n")
+            link(pr_url)
```
SuperSandro2000, line 331:
```suggestion
            link(f"https://github.com/NixOS/nixpkgs/pull/{pr}")
```

booxter: Rationale for this suggestion being?.. Either way seems fine, so unless there's a strong reason, I'd stick to status-quo.

pbsds: To the python virtual stack machine these two are nearly identical, but the former preserves `git blame` and may in the future see re-use

PR #533, opened 2025-07-17
Add --pr-json flag

Defelo:
This would allow tools like [nixpkgs-review-gha](https://github.com/Defelo/nixpkgs-review-gha), which start multiple instances of nixpkgs-review on different machines and combine their results, to ensure that all instances operate on the same PR commits, even if someone pushes to the PR branch while some instances are already running and others are still starting.

The idea is to fetch the PR data via the `pulls/{number}` endpoint once and provide the resulting JSON object via the ~~`NIXPKGS_REVIEW_PULL_REQUEST_JSON` environment variable~~ `--pr-json` flag to all nixpkgs-review instances. This ensures that no API calls to the `pulls/{number}` endpoint are made at all by nixpkgs-review.

See https://github.com/Defelo/nixpkgs-review-gha/pull/31 for the nixpkgs-review-gha feature for which this would be useful.

### DIFF
--- a/nixpkgs_review/cli/__init__.py
+++ b/nixpkgs_review/cli/__init__.py
@@ -1,4 +1,5 @@
 import argparse
+import json
 import os
 import re
 import subprocess
@@ -30,6 +31,14 @@ def regex_type(s: str) -> Pattern[str]:
         return re.compile(s)
     except re.error as e:
         msg = f"'{s}' is not a valid regex: {e}"
+        raise argparse.ArgumentTypeError(msg) from e
+
+
+def json_type(s: str) -> Any:
+    try:
+        return json.loads(s)
+    except json.JSONDecodeError as e:
+        msg = f"'{s}' is not a valid JSON object: {e}"
         raise argparse.ArgumentTypeError(msg) from e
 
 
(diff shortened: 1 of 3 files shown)


SuperSandro2000 (commented):
Do we have such variables somewhere documented? Otherwise LGTM

nixpkgs_review/github.py
```diff
@@ -114,6 +115,12 @@ def graphql(self, query: str) -> dict[str, Any]:
 
     def pull_request(self, number: int) -> Any:
         "Get a pull request"
+
+        if pr_json := os.environ.get("NIXPKGS_REVIEW_PULL_REQUEST_JSON"):
```
MattSturgeon, line 119: I won't block on this, but it'd be nice to also have a CLI argument in addition to the environment variable.

MattSturgeon (approved):
Nice!

Defelo:
I decided to remove the environment variable and instead simply use cli args for this, since we don't seem to be using env vars for any other user input, and cli args are already documented in `--help`. This approach also makes it easier to provide json objects for more than one PR, if multiple PRs are reviewed at once.

Issue #476, opened 2025-03-10
`--eval github` sometimes reports `urllib.error.HTTPError: HTTP Error 410: Gone`


FliegendeWurst:
For example, when trying to review https://github.com/NixOS/nixpkgs/pull/362297.

I would assume the error occurs because the eval artifact disappeared.

I think the error should be reported without dumping a stacktrace on the user. Ideally it would then evaluate locally.

```
Traceback (most recent call last):
  File "/nix/store/4av055vaj05smk7spcskwkpdhy90qgam-nixpkgs-review-3.0.1/bin/.nixpkgs-review-wrapped", line 9, in <module>
    sys.exit(main())
             ^^^^^^
  File "/nix/store/4av055vaj05smk7spcskwkpdhy90qgam-nixpkgs-review-3.0.1/lib/python3.12/site-packages/nixpkgs_review/__init__.py", line 10, in main
    cli.main(command, args)
  File "/nix/store/4av055vaj05smk7spcskwkpdhy90qgam-nixpkgs-review-3.0.1/lib/python3.12/site-packages/nixpkgs_review/cli/__init__.py", line 341, in main
    return cast(str, args.func(args))
                     ^^^^^^^^^^^^^^^
  File "/nix/store/4av055vaj05smk7spcskwkpdhy90qgam-nixpkgs-review-3.0.1/lib/python3.12/site-packages/nixpkgs_review/cli/pr.py", line 115, in pr_command
    contexts.append((pr, builddir.path, review.build_pr(pr)))
                                        ^^^^^^^^^^^^^^^^^^^
  File "/nix/store/4av055vaj05smk7spcskwkpdhy90qgam-nixpkgs-review-3.0.1/lib/python3.12/site-packages/nixpkgs_review/review.py", line 302, in build_pr
    packages_per_system = self.github_client.get_github_action_eval_result(pr)
                          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/nix/store/4av055vaj05smk7spcskwkpdhy90qgam-nixpkgs-review-3.0.1/lib/python3.12/site-packages/nixpkgs_review/github.py", line 183, in get_github_action_eval_result
    changed_paths: Any = self.get_json_from_artifact(
                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #546, opened 2025-08-02
fix(github): handle expired artifacts with custom exception

khaneliman:
Throw a custom exception when we a get 410 on fetching CI results and handle accordingly in the review process to fallback to local evaluation. 

Closes https://github.com/Mic92/nixpkgs-review/issues/476

Dependent on https://github.com/Mic92/nixpkgs-review/pull/545 for CI, pulled these changes into separate branch so it doesn't hold that one up. 

---------------------------------------------------

Just saw this in the associated issue

> We should treat 410 the same way as 404.

Should i make the exception raised on `404`, as well? My understanding was that 404 waits for a retry loop so that you can wait for an evaluation to finish. But, here we had an evaluation that did finish and we know it won't because it just expired so we just force a local retry.

### DIFF
--- a/nixpkgs_review/errors.py
+++ b/nixpkgs_review/errors.py
@@ -1,2 +1,6 @@
 class NixpkgsReviewError(Exception):
     """Base class for exceptions in this module."""
+
+
+class ArtifactExpiredError(NixpkgsReviewError):
+    """Raised when GitHub artifacts have expired or been removed."""
--- a/nixpkgs_review/github.py
+++ b/nixpkgs_review/github.py
@@ -7,8 +7,10 @@ import urllib.request
 import zipfile
 from http.client import HTTPMessage
 from pathlib import Path
+from textwrap import dedent
 from typing import IO, Any, override
 
+from .errors import ArtifactExpiredError
 from .utils import System, warn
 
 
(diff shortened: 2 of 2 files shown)


GaetanLepage (approved):
This is a good idea. Diff LGTM

nixpkgs_review/review.py
```diff
@@ -423,29 +423,46 @@ def build_pr(self, pr_number: int) -> dict[System, list[Attr]]:
...
+                        waiting_time_s += timeout_s
+                        print(".", end="")
+                        sys.stdout.flush()
+                        time.sleep(timeout_s)
+                        try:
+                            packages_per_system = (
+                                self.github_client.get_github_action_eval_result(pr)
+                            )
+                        except ArtifactExpiredError as e:
+                            warn(f"\n{e}. Falling back to local evaluation.")
```
SuperSandro2000, line 446: Does this respect --eval github and not fallback to local then?

SuperSandro2000: I wouldn't want eval to run locally on my laptop as that most likely will OOM kill.

khaneliman:
If we just want a better message, this is what it looks like atm.

```bash
-> Fetching eval results from GitHub actions
https://github.com/NixOS/nixpkgs/pull/391575 failed to build:
GitHub artifact 3034442382 has expired or been removed
    * "try passing --eval local"
    * "try re-running GitHub CI"
```

PR #556, opened 2025-08-29
Force local evaluation when a non-empy --extra-nixpkgs-config is provided

GaetanLepage:
When the user provides a non-empty `--extra-nixpkgs-config`, the evaluation results from the GitHub action might not be valid.

In such cases, we should force local evaluation.
For instance, when `--extra-nixpkgs-config "{ allowUnfree = true; cudaSupport = true; }"` is provided, `nixpkgs-review` does not rebuild the correct package set.

cc @prusnak

## Summary by CodeRabbit

* **Bug Fixes**
  * GitHub-based evaluation is now automatically skipped when a non-default nixpkgs config is supplied via --extra-nixpkgs-config.
  * A clear message is shown to indicate fallback to local evaluation.
  * Ensures evaluation uses the specified custom configuration.
  * Default behavior unchanged when the default (empty) config is used.

### DIFF
--- a/nixpkgs_review/review.py
+++ b/nixpkgs_review/review.py
@@ -425,6 +425,16 @@ class Review:
             self._display_pr_info(pr, pr_number)
 
         packages_per_system: dict[System, set[str]] | None = None
+
+        # GHA evaluation only evaluates nixpkgs with an empty config.
+        # Its results should not be used when a non-default nixpkgs config is requested
+        normalized_config = self.extra_nixpkgs_config.replace(" ", "")
+        if self.use_github_eval and (normalized_config != "{}"):
+            print(
+                "Non-default --extra-nixpkgs-config provided. Falling back to local evaluation"
+            )
+            self.use_github_eval = False
+
         if self.use_github_eval:
             assert all(system in PLATFORMS for system in self.systems)
             print("-> Fetching eval results from GitHub actions")


pbsds:
This makes it impossible to use the github eval, even when `--eval github` is added explicitly

GaetanLepage:
> This makes it impossible to use the github eval, even when `--eval github` is added explicitly

But are you providing a non-empty `--extra-nixpkgs-config`?

pbsds:
Yes

```
--extra-nixpkgs-config '{ allowInsecurePredicate = x: true; }' --eval github
```

GaetanLepage:
> Yes
> 
> ```
> --extra-nixpkgs-config '{ allowInsecurePredicate = x: true; }' --eval github
> ```

Ok, so in this case, you would still want to use the GHA eval results.
Maybe what we should do is letting the user override the evaluation and still fetch the GHA results if `--eval github` is explicitly provided. Would that be fine @pbsds ?

Mic92:
I think we might have to leave it up to the user to specify `--eval local` manually.

GaetanLepage:
> I think we might have to leave it up to the user to specify `--eval local` manually.

So basically reverting this PR?

Wouldn't the following be better though?

| `--extra-nixpkgs-config` | `--eval` | Behavior |
|----|-------|-----|
| `*` | `local` | local eval |
| `{}` / unspecified | unspecified \| `github` | GHA eval (classic behavior) |
| custom | unspecified | **local** + warning (implemented by this PR) |
| custom | `github` | **github** + warning "Forcing GHA eval fetching, but we cannot guarantee the result will make sense" (TODO) |

We could implement the last line in a separate PR. This would let the user more freedom while still preventing him from shooting himself in the foot involuntarily.

PR #562, opened 2025-09-05
Support --include-pr to merge additional PR into reviewed worktree

booxter:
Sometimes a PR can be fully validated only with another unmerged PR. In this scenario, one either has to wait for the first PR to merge, or construct an alternative test path that is not using nixpkgs-review.

This patch solves it by adding support for --include-pr option that, if used, will merge additional PRs into the worktree under test.

The final report will list the new option to avoid confusion or misleading the reader as to the final state under test.

The option can be passed multiple times to include multiple PRs. The order of merges of the requested PRs is retained.

### DIFF
--- a/README.md
+++ b/README.md
@@ -444,6 +444,16 @@ either locally or through the remote builder protocol.
 $ nixpkgs-review pr --systems aarch64-linux 98734
 ```
 
+## Review changes with additional PRs included
+
+Sometimes a pull request depends on changes in another pull request to pass. To
+merge additional pull requests when reviewing a pull request, use the
+`--include-pr` flag multiple times:
+
+```console
+$ nixpkgs-review pr 12345 --include-pr 67890 --include-pr 98734
+```
+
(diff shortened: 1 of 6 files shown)


booxter:
@Mic92 before I spend time polishing this, please sanity check if such a feature would be acceptable. Thanks.

Mic92:
I guess makes sense for some ryantm pull requests.

booxter:
Example report here: https://github.com/NixOS/nixpkgs/pull/511235#issuecomment-4274752064

booxter:
@coderabbitai review

booxter:
@Mic92 this is ready for review

booxter:
Rebased and also explicitly disabled commit signing for merge commits that it creates (these commits are transient and temporary, so signatures are worthless; plus signing may also require interactive action from the user if their SSH key is e.g. guarded by an agent that requires a human action like Touch ID or password to unlock).

PR #570, opened 2025-10-17
Add --no-exit-status flag

Defelo:
## Summary by CodeRabbit

* **New Features**
  * Added `--no-exit-status` CLI flag to modify program exit behavior when package builds fail.

### DIFF
--- a/nixpkgs_review/cli/__init__.py
+++ b/nixpkgs_review/cli/__init__.py
@@ -200,6 +200,11 @@ def common_flags() -> list[CommonFlag]:
             help="Only evaluate and build without executing nix-shell",
         ),
         CommonFlag(
+            "--no-exit-status",
+            action="store_true",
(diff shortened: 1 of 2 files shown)


SuperSandro2000:
Can't this be simple replaced by `|| true`?

Defelo:
> Can't this be simple replaced by `|| true`?

That's what we used before in nixpkgs-review-gha, see the discussion in https://github.com/Defelo/nixpkgs-review-gha/issues/59. The problem is that we still want nixpkgs-review to fail if an error other than a package build failure occurs (e.g. fetching the eval results from gha failed).

Currently we simply ignore any error in nixpkgs-review using `|| true`, which in case of an error causes a later step to fail because the `report.json` is missing. Of course we could also check whether this file exists immediately after running `nixpkgs-review ... || true`, but I feel like an explicit flag would be a cleaner and more reliable solution.

PR #575, opened 2025-11-16
Isolate git operations with a new repository

### LINKED ISSUE #573: `git fetch` messes up shallow clones
I've got a shallow clone of `nixpkgs` together with a script to let me "`git fetch`" to it properly. Long story short, its `master` tracks NixOS/nixpkgs's `master`, its depth is neither an integer nor fixed, and when I make a branch to do something `master` will usually end up advancing by the time I'm ready to run `nixpkgs-review`

```
      o <---- master (advances with a cronjob)
     /|\
    / | \          o <- my-topic-branch
   / /|  \/-------/
  / | |\  \
 /  | | |  \
~~~~~~~~~~~~~ <- commits listed in .git/shallow (widened over time by cronjob)
```

`nixpkgs-review` puts my workflow in peril in a couple of ways:

* It does `git fetch --depth=<n>` if it sees the repository is shallow. While it does take care to use its own namespace of refs, `.git/shallow` is *not* namespaced. The `--depth=<n>` *beheads* my history, disconnecting `master` from `my-topic-branch`. This *causes* the dreaded "refusing to merge unrelated histories" error, even if my `master` was already up to date (that is, if a `fetch` was not necessary) and even if my repository otherwise contained enough history to merge `master` and `my-topic-branch`.

  That is, `nixpkgs-review` turns my repository into the following and then dies about it

  ```
        o <---- master (advances with a cronjob)
        ~ <- new shallow boundary due to --depth=<n>
                     
            o <- my-topic-branch
           / \
          /   \
         ~~~~~~~ <- partial old contents of .git/shallow
  ```

  Not only does `nixpkgs-review` fail, but a large chunk of the repository has become unreachable. I can manually fix it by deleting the offending new commits from `.git/shallow`, but in the mean time a `git gc` could theoretically remove some of the unreachable commits and prune the now-extraneous entries from `.git/shallow` and make the situation permanent.

  TL;DR: `git fetch --depth=<n>` is logically a destructive operation. `nixpkgs-review` shouldn't be doing it.
* It does `git fetch` at all. `git fetch` (at least, as of `git-2.50.1`) simply doesn't handle shallow repositories very well. Even if `nixpkgs-review` were patched not to pass `--depth=<n>` to `git fetch`, if NixOS/nixpkgs has advanced beyond my `master` and `nixpkgs-review` does a `git fetch` of the new tip commit, then that `git fetch` may pull in a line of commits that goes around my shallow boundary and thus start pulling down the whole history of nixpkgs.

  ```
        o <---- NixOS/nixpkgs's master
       /|\
      / | \
     /  o <----- my master
    /  /|\  \
   /  / | \  \
  |  ~~~~~~ <- my .git/shallow
  |
  v to rest of nixpkgs history (huge)
    (to add insult to injury git fetch is incredibly slow at negotiating which commits it wants in this case)
  ```

  This is much less of a problem; were this to happen I could just interrupt the fetch and run my own script to update my repository properly before rerunning `nixpkgs-review`.

All that said, I think `nixpkgs-review` should consider doing some or all of the following:
* Provide an option to override the shallow repository detection and make sure `--depth` is never passed to `git fetch`.
  * Actually, I can't think of a situation where *anyone* would benefit from `nixpkgs-review` passing `--depth` to `git fetch`. Maybe this code should be off by default or even removed?
* Provide an option to make sure `git fetch` never happens at all. If no new commits need to be fetched anyway, then `nixpkgs-review` should simply create a ref without using `git fetch`. If new commits do need to be fetched, `nixpkgs-review` could either let the user specify a custom command or simply error out.
* Change the comment in the README about shallow clones. The problem with shallow clones is not so much insufficient history being available to merge topic branches into master, but rather the standard `git` tools simply don't know how to bring a shallow repository up to date without either breaking the history or breaking the shallowness (and current `nixpkgs-review` chooses to break the history).

howtonotwin:
Shallow fetch operations permanently change the shallow configuration of the repository they run in. Using a temporary worktree does not protect the user's repository against such modifications. Switch to using a new separate repository to provide appropriate protection. Fixes #573.

Set up the alternates mechanism to have the new repository share storage with the original. This cannot be done automatically by `git-clone` since `git-clone` refuses to use a shallow repository as an alternate object store (for good reason!). Manually create the new repository and provide it with a copy of the original's shallow configuration to make this use of alternates correct.

On the way I ended up doing a bit of refactoring re: the handling of `nixpkgs-review wip`.

This PR is a little rough around the edges, but it does seem to work and not bork my repository anymore.

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,13 +60,11 @@ First, change to your local nixpkgs repository directory, i.e.:
 cd ~/git/nixpkgs
 ```
 
-If you've shallow cloned nixpkgs (`git clone --depth`), `nixpkgs-review` may be
-unable to perform merges due to missing merge base commits. Reclone nixpkgs
-without the `--depth` flag.
-
-Note that your local checkout git will not be affected by `nixpkgs-review`,
-since it will use [git-worktree](https://git-scm.com/docs/git-worktree) to
-perform fast checkouts.
+Note that `nixpkgs-review` will not leave any effects in your local Git
+repository (if it does, it's a bug!). It does its work in a temporary repository
+separate from your own. It uses Git's
+["alternates"](https://git-scm.com/docs/gitrepository-layout) mechanism to
+minimize transfers.
 
 Then run `nixpkgs-review` by providing the pull request number…
 
@@ -153,6 +151,18 @@ nix-shell> nixpkgs-review post-result
 # Review-comments can also be shown
 nix-shell> nixpkgs-review comments
 ```
+
+## Caveats for shallow clones
+
+If you've shallow cloned nixpkgs (`git clone --depth`), `nixpkgs-review` may be
+unable to perform merges due to missing merge base commits. The simplest fix for
+this is, as you might imagine, recloning without the `--depth` flag. Even in
+cases where your shallow cuts do not interfere with merging, it is possible that
+the latest upstream commits provide a new path "around" your shallow boundary,
+in which case `git fetch` will (very, very slowly) more-or-less unshallow your
+repository. This is not a problem in `nixpkgs-review`, but a limitation of Git.
+It is possible to get around this by using a more intelligent fetching program
+to update your repository before running `nixpkgs-review`.
 
 ## Optional tools
 
--- a/nixpkgs_review/builddir.py
+++ b/nixpkgs_review/builddir.py
@@ -6,7 +6,6 @@ from pathlib import Path
 from tempfile import TemporaryDirectory
 from typing import TYPE_CHECKING, Self
 
-from . import git
 from .overlay import Overlay
 from .utils import warn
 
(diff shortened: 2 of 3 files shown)


SuperSandro2000:
> The review sandbox conversely doesn't inherit configuration or references from the original repository.

Which can be a problem and as you wrote further down we don't want to open a can of worms propagating those. 

> because the standard git tools have generally poor support for them anyway

Which is bad

> requires low-level Git surgery and should be automated

Low level surgery could easily break on an update because it is low level things which we should avoid using if possible. 

----

I have a hard time seeing that this outweighs the downsides. We make the logic even more complicated for something very little people use and testing it is also not easy. If you effectively want to contribute to nixpkgs you need a pretty full clone anyway as otherwise rebasing on staging or doing a backport is not possible. Also that every git command needs to be handled very carefully makes this very easy to create in the future with every change related to git commands which is not possible to be checked by everyone.

howtonotwin:
I mostly wrote this to have a `nixpkgs-review` that works for me (and to convince myself that it was at all possible to do what nixpkgs-review wants to do "correctly"). I don't mind if this PR doesn't get merged/is closed.

I do think that mainline (which already special cases shallow repositories) should at least warn/require a flag/document/do *something* to clarify that it is *destructive* (not just liable to error out) when run on a shallow clone.

Issue #244, opened 2021-12-22
Compare to build of base.


kevincox:
Right now nixpkgs-review simply tells you about the build status at the selected commit. it would be great if it could compare this to the base state. For example "2 failing -> failing" is very different than "2 building -> failing".

SuperSandro2000:
https://github.com/SuperSandro2000/nixpkgs-review-checks implements that in a hacky way.

PR #583, opened 2026-02-01
Add '--checkout base' option to build reference

raboof:
To more quickly check whether failures in the PR also already fail on the base/target branch.

The most primitive implementation of #244

### DIFF
--- a/nixpkgs_review/review.py
+++ b/nixpkgs_review/review.py
@@ -541,6 +541,12 @@ class Review:
             }
 
         if packages_per_system is None:
+            if self.checkout == CheckoutOption.BASE:
+                die(
+                    "--checkout base without --package/-p requires GitHub evaluation.\n"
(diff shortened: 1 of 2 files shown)


raboof:
ah yes, updated :)

nixpkgs_review/cli/__init__.py
```diff
@@ -66,14 +66,15 @@ def pr_flags(
     checkout_help = (
         "What to source checkout when building: "
         "`merge` will merge the pull request into the target branch, "
-        "while `commit` will checkout pull request as the user has committed it"
+        "while `commit` will checkout pull request as the user has committed it."
+        "`base` will build the base (target) branch, for reference."
```
Mic92: I think we should generate a warning if you don't use the ofborg evaluation here, because than you will never build anything.

PR #595, opened 2026-02-11
Use nix-eval-jobs in multi_system_eval

B4dM4n:
This is part one of three in my attempt to make local test evaluation feasible (see #397) by optimizing the way changed packages are instantiated/build.

Switching to `nix-eval-jobs` when evaluating the changed packages in `filter_packages`/`join_packages`/`nix_build` does not provide a performance benefit in itself, but it lays the foundation to performing bigger local evaluations, by allowing to limit memory consumption.

Because `nix-eval-jobs` outputs one line of JSON attributes per derivation, this commit changes how outputs are mapped to `nix.Attr`'s.

Before this, each output was mapped to a separate `nix.Attr` and subsequently reported as a separate package in the final report.

Now, one derivation is mapped to one `nix.Attr` with multiple outputs. Duplicate `nix.Attr`'s are now identified by their `drvPath`'s, to avoid comparing the `outputs` dictionary.  In the report, each `nix.Attr` is now only reported once, without mentioning the outputs. But the outputs are still symlinked in the `results` directory.

The report could be modified to include the outputs again, but I think listing them only pollutes the resulting package list without much benefit. Individual outputs can't be built or filtered and are inherently tied to their package/derivation.

### DIFF
--- a/nixpkgs_review/nix.py
+++ b/nixpkgs_review/nix.py
@@ -326,8 +326,8 @@ def multi_system_eval(
                 raise TypeError(msg)
             # Skip error entries from nix-eval-jobs (e.g. system-level
             # evaluation failures) which lack per-package attrPath.
+            # nix-eval-jobs already prints these errors to stderr.
             if "error" in raw_result:
-                warn(f"nix-eval-jobs: {raw_result['error']}")
                 continue
             eval_result: NixEvalProps = raw_result  # type: ignore[assignment]
             system = eval_result["attrPath"][0]


nixpkgs_review/nix.py
```diff
@@ -336,10 +320,16 @@ def multi_system_eval(
...
-            if not isinstance(eval_result, dict):
-                msg = f"Expected eval result to be a dict, got {type(eval_result)}"
+            raw_result: dict[str, object] = json.loads(line)
+            if not isinstance(raw_result, dict):
+                msg = f"Expected eval result to be a dict, got {type(raw_result)}"
                 raise TypeError(msg)
+            # Skip error entries from nix-eval-jobs (e.g. system-level
+            # evaluation failures) which lack per-package attrPath.
+            if "error" in raw_result:
+                warn(f"nix-eval-jobs: {raw_result['error']}")
```
B4dM4n: Printing the error should not be necessary. `nix-eval-jobs` has already written the error to stderr in color at this point.

PR #605, opened 2026-03-05
Don't evaluate packages in `nix build`

B4dM4n:
Re-evaluating all attributes can significantly increase the memory consumption of `nix build` by several GB during the entire runtime of the process and also can take some time for large reviews.

The required derivations are already known from `multi_system_eval`, so pass them directly to `nix build`.

Sadly, we can't pass derivations or store paths directly to `mkShell`/`nix-shell`, so we need to turn the required outputs to store paths with `builtins.storePath` and pass them to `mkShell`.

### DIFF
--- a/nixpkgs_review/nix.py
+++ b/nixpkgs_review/nix.py
@@ -24,7 +24,6 @@ class BuildConfig:
 
     allow: AllowedFeatures
     nix_path: str
-    local_system: str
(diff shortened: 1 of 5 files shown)


Mic92:
Landed in reworked form as #690 (instantiate in nix-eval-jobs, nix build --stdin, review shell without nix eval). Thanks!

PR #613, opened 2026-03-26
Support bare nixpkgs repositories

philiptaron:
# Supporting bare nixpkgs repositories

## The problem

I keep my nixpkgs checkout as a bare repo with worktrees:

```
~/nixpkgs/
├── .git/              (core.bare = true)
├── master/            (worktree)
├── staging/           (worktree)
├── pr-work/           (worktree)
└── ...
```

I learned this git workflow for large repos with many workstreams (like nixpkgs)! from @ericson2314. You clone once, set `core.bare = true`, and then `git worktree add` for each branch you're working on. No files are checked out at the root; the worktrees live as sibling
directories alongside `.git/`. It's pretty awesome.

Today, when I `cd ~/nixpkgs` and run `nixpkgs-review pr 12345`, it dies immediately with "Has to be executed from nixpkgs repository". The tool can't tell it's inside nixpkgs because there's no `nixos/release.nix` on disk at the root level -- it only exists inside the worktrees.

Running from inside a worktree (`cd ~/nixpkgs/master && nixpkgs-review pr ...`) works fine, but it's annoying to have to pick an arbitrary worktree just to launch a review that creates its own worktree anyway.

## What was broken

Two functions assumed a working tree exists:

1. **`find_nixpkgs_root()`** walks up from CWD looking for `nixos/release.nix` on the filesystem. In a bare repo there are no checked-out files, so it walks all the way to `/` and gives up.

2. **`resolve_git_dir()`** looks for a `.git` file or directory.  My layout has `.git/` as a directory so this one actually works for me, but a "true" bare clone (`git clone --bare`) has no `.git` entry at all -- the git database (HEAD, objects, refs) lives directly in the directory. That case would also fail.

## What this PR does

**`find_nixpkgs_root()`** now falls back to git when the filesystem walk fails. It asks `git rev-parse --is-bare-repository` whether we're in a bare repo, then checks `git cat-file -e HEAD:nixos/release.nix` to confirm the repo actually contains nixpkgs (not just any bare repo). If both pass, it returns CWD as the root.

**`resolve_git_dir()`** gets a similar fallback: when there's no `.git` on disk, it tries `git rev-parse --git-dir`, which works in true bare clones where the git database is the directory itself.

**`wip` is guarded** with an early exit in bare repos. The `wip` command captures the dirty working tree via `git diff`, but a bare repo has no working tree to diff.  Rather than letting it silently produce an empty diff and exit, it now tells you to use `pr` or `rev` instead.

The non-bare case is unchanged -- the filesystem walk still runs first, so there's no subprocess overhead for the common case.

## Refactoring

Along the way, `find_nixpkgs_root`, `resolve_git_dir`, `fetch_refs`, and friends were extracted from `buildenv.py` and `review.py` into a new `nixpkgs.py` module, since they're all about locating and interacting with the nixpkgs git repo rather than about the build environment or the review workflow.

The test suite also got a session-scoped fixture that sets `GIT_CONFIG_GLOBAL` and `GIT_CONFIG_SYSTEM` to `/dev/null`, so the user's git config (things like `commit.gpgsign = true`) can't break test commits. Previously -- at least on my system -- this required passing those env vars on the command line every time I ran `pytest`.

## AI-assisted

As you likely can tell, my friend Claude helped with this implementation. I've reviewed it and think it's pretty slop-free, but please let me know your own tastes on the matter.

### DIFF
--- a/nixpkgs_review/buildenv.py
+++ b/nixpkgs_review/buildenv.py
@@ -6,20 +6,11 @@ from pathlib import Path
 from tempfile import NamedTemporaryFile
 from typing import TYPE_CHECKING
 
+from .nixpkgs import find_nixpkgs_root
 from .utils import die
 
 if TYPE_CHECKING:
     import types
-
-
-def find_nixpkgs_root() -> Path | None:
-    root_path = Path.cwd()
-    while True:
-        if (root_path / "nixos" / "release.nix").exists():
-            return root_path
-        if root_path == root_path.parent:
-            return None
-        root_path = root_path.parent
 
 
 class Buildenv:
--- a/nixpkgs_review/cli/wip.py
+++ b/nixpkgs_review/cli/wip.py
@@ -6,12 +6,14 @@ from typing import TYPE_CHECKING
 from nixpkgs_review import git
 from nixpkgs_review.allow import AllowedFeatures
 from nixpkgs_review.buildenv import Buildenv
+from nixpkgs_review.nixpkgs import is_bare_repository
 from nixpkgs_review.review import (
     LocalRevisionTarget,
     ReviewAction,
     build_config_from_args,
     review_local_revision,
 )
+from nixpkgs_review.utils import die
 
 if TYPE_CHECKING:
     import argparse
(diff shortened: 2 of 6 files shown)


philiptaron:
Awesome, thank you.

PR #620, opened 2026-04-09
update instructions to use without installing

gamebeaker:
The example shows how to add flags to the command without installing it.

I added `--post-result` as i think that it is mostly used for this purpose.

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,7 +28,7 @@ automatically builds packages changed in the pull requests.
 To use it without installing it, use:
 
 ```console
-$ nix run 'nixpkgs#nixpkgs-review'
+$ nix run 'nixpkgs#nixpkgs-review' -- pr <PR number>
 ```
 
 To run it from the git repository:


README.md
```diff
@@ -28,7 +28,7 @@ automatically builds packages changed in the pull requests.
 To use it without installing it, use:
 
 ```console
-$ nix run 'nixpkgs#nixpkgs-review'
+$ nix run nixpkgs#nixpkgs-review -- pr --post-result <PR number>
```
SuperSandro2000: I think the quotes are required for some shells.

SuperSandro2000:
```suggestion
$ nix run nixpkgs#nixpkgs-review -- pr <PR number>
```
or does the interactive shell not open?

gamebeaker:
> I think the quotes are required for some shells.

@SuperSandro2000 i get this error with the quotes (that is the reason why i made the PR):
image

The problem is the flag `--post-result` without it there is no problem:
image

That is why i included it in the example.

figsoda:
> i get this error with the quotes (that is the reason why i made the PR):

This is because `--post-result` is passed to `nix` instead of `nixpkgs-review`. You should do this instead:

```bash
nix run 'nixpkgs#nixpkgs-review' -- pr --post-result 508042
```

Does removing the single quotes fix the issue for you? If so, I'm not sure why that would be the case.

gamebeaker:
@figsoda you are right. It didn't have anything todo with the quotes.
I just got it to work and thought the quotes were the problem.
Why does the following command work:
```
nix run 'nixpkgs#nixpkgs-review' pr 508042
```
and this one doesn't?
```
nix run 'nixpkgs#nixpkgs-review' pr --post-result 508042
```
> This is because --post-result is passed to nix instead of nixpkgs-review

wouldn't that also mean that in the first example `pr 508042` would have been passed to nix but it still executes nixpkgs-review with the correct flags?! and if there is a flag `--post-result` more it suddenly throws an error.

figsoda:
> Why does the following command work:

`nix run` only accepts one positional argument, so `pr` and `508042` are unambiguous and can only be passed to `nixpkgs-review`

`--post-result` starts with `-`, and nix's [argument parsing logic](https://github.com/NixOS/nix/blob/14ffc1787182b8702910788aea02bd5804afb32e/src/libutil/args.cc#L344) takes it as its own flag unless it's after `--`

figsoda (approved):

PR #623, opened 2026-04-18
fix: avoid no-url-literals warning

booxter:
Nix 2.34+ warns:

```
experimental feature 'no-url-literals' has been stabilized and
renamed; use 'lint-url-literals = fatal' setting instead
```

Use lint-url-literals instead of the deprecated no-url-literals experimental feature. Older Nix releases ignore the option, so URL literals remain allowed there.

### DIFF
--- a/nixpkgs_review/nix.py
+++ b/nixpkgs_review/nix.py
@@ -86,7 +86,8 @@ REVIEW_SHELL: Final[str] = str(ROOT.joinpath("nix/review-shell.nix"))
 def _nix_common_flags(allow: AllowedFeatures, nix_path: str) -> list[str]:
     return [
         "--extra-experimental-features",
-        "nix-command" if allow.url_literals else "nix-command no-url-literals",
+        "nix-command",
+        *([] if allow.url_literals else ["--option", "lint-url-literals", "fatal"]),
         "--nix-path",
         nix_path,
         "--allow-import-from-derivation"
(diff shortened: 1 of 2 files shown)


SuperSandro2000:
Why do we even set that option? Could we just omit it entirely?

figsoda:
url literals are not allowed in nixpkgs, and it lets `nixpkgs-review wip` or `nixpkgs-review rev HEAD` catch it locally before ci

nixpkgs_review/allow.py
```diff
@@ -1,5 +1,24 @@
 from __future__ import annotations
 
+import functools
+import re
+import subprocess
+
+
+@functools.lru_cache(maxsize=1)
```
figsoda: this is a lot of logic for a small feature like this

i think we can just stick to `--option lint-url-literals fatal`, which will just be an `unknown setting` warning for older nix versions

xokdvium: That was my thinking too, I suppose nix could also just silence the XP feature warning if the option is enabled to be less annoying.

booxter: @figsoda To clarify, do you suggest that we disable the `no-url-literals` feature for users of older Nix versions? I didn't want to regress on them, but if you insist, I can oblige. Let me know.

figsoda: yes, I suggest that we use `--option lint-url-literals fatal` for all nix versions, which would allow url literals on older nix versions

figsoda (approved):
tested with an updated flake.lock

PR #636, opened 2026-05-09
Omit `<details>` if 2 or fewer error logs

pbsds:
saves a click for leaf packages, which make up the majority of packages. This takes no/little additional space

tests:

* https://github.com/NixOS/nixpkgs/pull/508740#issuecomment-4411029923
* https://github.com/NixOS/nixpkgs/pull/516867#issuecomment-4412501168

### DIFF
--- a/nixpkgs_review/report.py
+++ b/nixpkgs_review/report.py
@@ -106,7 +106,7 @@ def remove_ansi_escape_sequences(text: str) -> str:
 
 
 def html_logs_section(logs_dir: Path, packages: list[Attr], system: str) -> str:
-    res = ""
+    logs = []
     seen_tails = set()
     for pkg in packages:
         if tail := html.escape(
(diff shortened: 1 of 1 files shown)


nixpkgs_review/report.py
```diff
@@ -106,24 +106,24 @@ def remove_ansi_escape_sequences(text: str) -> str:
...
-            res += f"<details>\n<summary>{html.escape(pkg.name)}</summary>\n<pre>{tail}</pre>\n</details>\n"
+            logs.append(f"<details>\n<summary>{html.escape(pkg.name)}</summary>\n<pre>{tail}</pre>\n</details>")
             seen_tails.add(tail)
-    if res:
-        res += "</details>\n"
-    return res
+    if not logs:
+        return ""
+    elif len(logs) <= 2:
+        return f"\n---\n{"\n".join(logs)}\n"
```
pbsds: this could be set to 1, which makes for no extra height

SuperSandro2000 (approved):
LGTM

nixpkgs_review/report.py
SuperSandro2000: I think 2 is very fine, maybe even 3, but lets see how it goes.

PR #646, opened 2026-05-26
fix: build all outputs of each package

### LINKED ISSUE #645: Build shows as failed when one output of a package is already present and another is not
What I did: tried running `nix-review pr` when one output of a package was already built and another was not. (In my particular case, the output that was "already built", the `out` output, was in fact substituted; the output that was not already built, and was not substitutable, was the `man` output.)

What I expected would happen: nixpkgs-review either doesn't care about the unbuilt output, or it builds it.

What actually happens: the entire package shows as "failed to build".

This looks like it was likely caused by [8ed3acff617036e0deadb8f65e86290a9b4eec9e](https://github.com/Mic92/nixpkgs-review/commit/8ed3acff617036e0deadb8f65e86290a9b4eec9e#diff-f5b281a937846b7bc507e9a95a782fac5b1bcfcec1ea1a2348abc32b0282038eR48) (line 48, along with the related changes).

<s>Not sure what the fix should be, though...</s>

mdaniels5757:
This fixes #645 by simply building all outputs of each package.

### DIFF
--- a/nixpkgs_review/nix/review-shell.nix
+++ b/nixpkgs_review/nix/review-shell.nix
@@ -26,7 +26,13 @@ let
         config = nixpkgs-config;
       };
     in
-    map (attrString: lib.attrByPath (lib.splitString "." attrString) null system-pkg) system-attrs;
+    map (
+      attrString:
+      let
+        attr = lib.attrByPath (lib.splitString "." attrString) null system-pkg;
+      in
+      attr.all or attr
+    ) system-attrs;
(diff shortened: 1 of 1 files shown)


GaetanLepage (approved):
Nice! I was bitten by this recently.

@Mic92 can you have a look a the fix too pls?

SuperSandro2000:
How hard would it be to add a test case for this?

mdaniels5757:
Rather hard, I think: it would need to ensure that the nix store has some outputs of a package but not others, and it would need to be a different outpath every time (so builds of the remaining outputs don't cause a problem on the next run).

